PR #2, opened 2018-01-23
Rework hal interface

Zaltora:
Since i want to use lvgl, i want to use some drivers too.
The original lvgl system use macro and makes some configurations rigid.
For example:
CS is hard coded and can't be managed by SPI.
I2C macro manually start, stop... 
Concurrent system problematic when a bus is used. 

I totally rework the abstraction and create a  HAL LVGL API for common usage (it is WIP ) : 
```
void lv_delay_us(uint32_t us);
void lv_delay_ms(uint32_t ms);
void lv_gpio_write(uint16_t pin, uint8_t val);
uint8_t lv_gpio_read(uint16_t pin);
int lv_i2c_write(void* i2c_dev, const uint8_t* reg, const void* data_out, uint16_t datalen);
int lv_i2c_read(void* i2c_dev, const uint8_t* reg, void* data_in, uint16_t datalen);
....
```
When people want use lvgl drivers, they just need to set function in "lv_drv_conf.c" and respect documentation.
The library use lvgl HAL system to use peripherals. Peripheral must be initialized by the app user. 
Current drivers need be rework to add a device descriptor that will contain pin information, bus, ...

The ssd1306 library work well. (tested with SSD1306/SH1106 SPI3 protocol).
I suggest to open a branch (the time to port  all drivers to new API).


kisvegabor:
Hi,

The only reason for I used macros instead of functions that I would like to keep the call of operations  fast. For example sometimes only direct port pin manipulations are required, like `PORTB.3 = 1`. 
I suggest to make possible the use of inline functions. In this case only a template header file need to be provided and user can add its 'c' files.

Zaltora:
User can do it. Add inline function in the c file has the same effect ?  (i do it). You want i add them for all functions by default ?
You prefer the HAL source into the app user ? not in the directory above ? 
My configuration:
    - component.mk
    - lv_conf.h
    - lv_drv_conf.c
    - lv_drv_conf.h
    - lv_drivers/
    - lv_utils/
    - lvgl/

kisvegabor:
As I know inline functions should be in the header files because when the compiler compiles a 'c' file it need to see the inline function directly not only the declarations.

Zaltora:
Interesting. I never thought about this details.
I see a solution. In the header, you add "extern inline", in the source, you put inline.
(Maybe i need to add this in lvgl for tick and handler functions ?)

[Answer](https://stackoverflow.com/questions/5057021/why-are-c-inline-functions-in-the-header#5057065)
[Details](http://www.greenend.org.uk/rjk/tech/inline.html)

kisvegabor:
To be honest it's not crystal clear to me. However inline functions in the header files is surely work. As I see it is up to the user how to write his functions. For example in lv_drv_conf.h:

```c
static inline void my_func(void)
{

}
```
or

```c
void my_func(void);
```
and declare it in the C file

Zaltora:
Okay, 
I will rename the source file with "opt" to indicate it is optional, and the user need just the "lv_drv_conf_templ.h". It is good like this ?

The inline subject was interesting so i got more clue about it:
[history explanation](https://stackoverflow.com/questions/216510/extern-inline#216546)
[other explanation](http://m68hc11.serveftp.org/inline-1.php)

What did you think with the global HAL API system ?

kisvegabor:
Hi,

According to these lines we can use `extern inline`:

> This combination of inline and extern has almost the effect of a macro. The way to use it is to put a function definition in a header file with these keywords, and put another copy of the definition (lacking inline and extern) in a library file. The definition in the header file will cause most calls to the function to be inlined. If any uses of the function remain, they will refer to the single copy in the library.

I like the idea to use functions instead of macros. 

A few questions:
1. Why pass e.g. `i2c_dev`? In my opinion the driver functions should know which I2C to use to reach the display. And the display will be always on this I2C so no point to specify the I2C device. 
2. `lv_spi_set_command` and `lv_spi_set_address` function names. For me `lv_disp_spi_wr` and `lv_disp_spi_wr` seems more flexible.
3. About gpio functions: What about function with exact names like: `lv_disp_gpio_set_cs()`. It also ensures faster access to the pin because you don't need to search the pin named in the argument. If you control CS with the SPI you can leave this function empty.  

So in general my vision is something similar as it was with macros originally: different "name space" for display and touch pad functions: `lv_disp_spi_...`, `lv_disp_gpio_...` `lv_disp_par_...` etc.

What do you think? I'm open to any constructive idea and update!

Zaltora:
The problem with question 1 and 3 is the same: The problem is to optimize the communication between MCU and devices as most as possible.
If we hard code device selection ( i2c_dev, cs pin, ...). We remove the ability to get more than one device of this type and the possibility to change some settings dynamically. In the current system, we suppose to get one screen and one input. Make it as parameter permit user to use multi-screen /multi input support. Dynamically, is the loss of performance huge? I am not sure. After all, for the performance, we use hardware peripherals. That why i put CS control pin by gpio optional in SSD1306 driver.
Other minor advantage, we don't need to set different namespace for different devices. When the HAL sources is written, only app user will define peripherals usage.
We can maybe make it as define  parameter, Swap between hard coded functions and dynamic instead (when use only one screen/input) but it can be heavy to write a driver.

For the question 2, Well , i didn't understand what you want to do but i am interesting. Can you details pls ? In my current platform, adress, dummy, and cmd are hardware SPI setting. In each different transaction, they are sent at start.

kisvegabor:
Hi,
I see your point. So your goal is to deal with multiple displays. Now there is no built-in support for multiple displays in the library, but according to my view it should be done by registering multiple display drivers (`lv_disp_drv_register()`). It is not supported yet and it is not easy to implement: different 'act_scr', different invalid area buffer, different resolution for each display? But is seems you would like to achieve this functionality on the driver level. Can you explain it in more detail?

About question 2: as I see every SPI command is an SPI write in the background. Then it is more general to specify only a write function instead of more higher level function. But lets focus on the multiple display part. It's is more relevant and highly determines the directions.

Zaltora:
You are right, lvgl don't support multi screen yet, but i think it is a great feature and we need to be prepared to this ( at least for the driver layer for the moment). We can do with hacky way to add multi-screen (like we discuss in an other [lvgl #73](https://github.com/littlevgl/lvgl/issues/73) but we need change driver, it can get bugs, and we can't get no more advantages than only a big screen.

Why multi-display: 
- More Display ( basic ^^) :  e.g.  One screen with some screen object (data visualizer ), an other with an other. A user can swap a window with a screen and keep what he want with other. Possibility to keep only one screen on for power saving usage instead of one big screen. ( Add 10 screen just because it is possible ^^ if you want too)
- Different display kind: e.g.  A TFT screen for fast interaction with user, a epaper for global informations system.
- Security/ Features : send display object to a screen to an other,  graphical computation optimization with screnn extend mode (e.g : 640*240 with two ILI9341 screen). Internal protection for concurrent task system,...

And multi Input ? 
- More screen, more input ^^. 
- Redundancy (maybe ?)
- Touch screen  with a Touch surface outside of the screen

Why do prepare it now in driver level:
This repository does not contain many drivers and can easily be updated. I do not doubt the possibility of lvgl to be very popular and drivers number will increase significantly. Add support for multi-screen will then be a suffering. 
In a broader sense, it would require a charter to design driver here.

The micro controllers are becoming more powerful to manage these functionality. e. g : esp 32 ( 520KRAM, 240MHz, 2 core, ... ), STM32H7 ( 400MHz, GPU, ...) . lvgl is good API for company. IT is got fundamentals need to design fast a interface ( Just missing the drag and drop software( like GLADE for GTK+) to put lvgl graphical object on a virtual screen but this feature can wait, it is a huge soft.

What did you think ? ( Sry for fault and misunderstanding). Design driver like independent object is good too you ? What we lost if we design like this, few memory call (or not, depend on hw) ?

kisvegabor:
Thanks for the detailed explanation! I agree, there can be cases where more displays are useful. (at least for fun :) ) So the scenario for multi screen  looks like this:
1. E.g. set double LV_HOR_RES for two displays 
2. The program runs and lvgl calls your `disp_flush` function. Here:
3. You cut the area and send the chunks  to the corresponding display

Am I correct? If yes how to pass a "half area" to a display? I suppose call the driver's flush function (e.g. `ssd1961_flush`). So the drivers flush function should know which display to refresh. In the current driver system I don't see how will it now that. My idea is to specify a function like: `lv_disp_spi_set_dev()` With this you can easily set a new device before calling the driver's flush function to choose the other display.

Zaltora:
I will base my answer on the drivers development . I am not familiar with the lvgl API yet. So I do not know if what I'm going to say is a huge job or not.

 - My idea

The features order to do to develop drivers correctly:
[ ] Multi Pixel by byte support  (1,2,4 bits)
[ ] Multi Display support (optionnal for lvgl now, but drivers will got a device descriptor)

With the first point, we don't need to use small buffer to save RAM for little screen (like SSD1306 driver ). And we can use internal lvgl VDB and multi features screen
For the multi display, I think we can no longer define macros for some parameters.
Therefore you have to register new parameters for "lvgl" driver:
the size (HEIGHT / WIDTH), the color, screen handle, ....
example:
~~~~~~~~~~~~~~{.c}
typedef struct _disp_drv_t {
    /*Write the internal buffer (VDB) to the display. 'lv_flush_ready()' has to be called when finished*/
    void (*disp_flush)(int32_t x1, int32_t y1, int32_t x2, int32_t y2, const lv_color_t * color_p, void* screen_handle);

    /*Fill an area with a color on the display*/
    void (*disp_fill)(int32_t x1, int32_t y1, int32_t x2, int32_t y2, lv_color_t color, void* screen_handle);

    /*Write pixel map (e.g. image) to the display*/
    void (*disp_map)(int32_t x1, int32_t y1, int32_t x2, int32_t y2, const lv_color_t * color_p, void* screen_handle);

#if USE_LV_GPU
    /*Blend two memories using opacity (GPU only)*/
    void (*mem_blend)(lv_color_t * dest, const lv_color_t * src, uint32_t length, lv_opa_t opa, void* screen_handle);

    /*Fill a memory with a color (GPU only)*/
    void (*mem_fill)(lv_color_t * dest, uint32_t length, lv_color_t color, void* screen_handle);
#endif
   void* screen_handle;
   uint16_t height;
   uint16_t width;
   lv_color_t color;
} lv_disp_drv_t;
~~~~~~~~~~~~~~
We should consider the screnn as objects that can be placed (TOP / BOT / LEFT / RIGHT / ...)

 - Your idea

1 - If the screen don't get the same size vertical and horizontal , it will be a problem with a "ghost area"
2 - Sry, didn't see where you want call function ? in handler loop ?
3 - Yeah cut an area ( from VDB or not) and send to the corresponding display. Just a question, you think we can use different VDB color sized if we use different kind of display?

Your idea to use `lv_disp_spi_set_dev()` is a good idea. I prefer to use a device handle. Imagine, one day, you put the full possibility to use multiple thread to refresh multi screen. Otherwise. Use handle guarantees a trouble-free use of the driver.

kisvegabor:
### Multi-display support

At the point 2 I mean: yes, in `lv_task_handler` lvgl will begin to redraw the invalid areas of the screen. As you know drawing happens in the VDB first and when it is ready VDB will be flushed to the display. However in this case you should have a 'main disp_flush' which 'distributes' the areas to the right display and call the disp_flush of the displays. Am I correct? 
If yes, then it is not required to use extra data in disp_drv (width, height) because the main disp_flush should know it. 
In the other case when lvgl would have a native support to multi-disp those data become relevant. In this case you can specify where (to which display) create a screen object.  

The simpler way is: you distribute the areas in the main disp_flush. It's drawback is you should have a VDB enough for all displays because they are treated as one in lvgl. And it is not possible to use different color depth. However I will look after how difficult is to implement a 'native' multi display support. It would enable the use of VDB for only the active (being refreshed) display and apply different color depths. Basically every display would be threated as unique. The main challenge will be to keep the simplicity for one displayed systems. 

### Device handler
I checked your SSD1306 driver in more detail. It's very good! You have done a very deep work!
I was missing one thing: how to set (assign) a  device handler to the display driver. According to my experiences not every driver library have the 'driver handler pointer' functionality. So I think it would be better to reference a driver with an enum like: *LV_I2C_1, LV_I2C_2 ... * and the `lv_i2c_write` should know how to write to a specific periphery. 
Anyway I'm convinced about the use of driver handlers directly in the functions too, as you suggested. Thank you for the endurance :)


### Function names
We had different thoughts about function names.
**delay** it's perfect
**gpio**  you specify only a *pin*. Typically a *port* and a *pin* name is required. (I know e.g. in ESP32 only pin numbers exists) For compatibility it should be extended with a port id/name too. Do you agree?
**i2c** I'm missing the device address from the parameters. And IMHO using 8 and 16 bit function is little bit confusing for the user. "Which one should I implement?" "Both are required?" What about a *cmd_len* parameter (1 or 2 byte) ?
**spi** for me it would be clearer and simpler to have an `lv_spi_wr(dev, array, lenght)` function instead of different function to set command, address etc. And the driver should create the 'array' with the corresponding data. SPI driver is just a driver to send/receive. Don't need to know anything about protocols. 


Sorry for being too long. I tried to answer every questionable point.

Zaltora:
### MULTI DISPLAY SUPPORT
 I don't no how you can do it. What i imagine. It is got the same effect as to get two screen on computer. Like we can drag object from on display to an other for example. Anyway, the changes need to be simpler. (Embedded style)
I agree with you but i can't help so much right now ^^.
I don't see how main_flush can work right now without handle, just with  X/Y coordinate ? That mean the driver need to select the good screen ?
In my point, i prefer an abstraction of screen usage. that mean lvgl will do the job and select screen to draw.
I think we should record the metadata also (size, color, ...) in the driver to facilitate the mixing of different screens types.

- Result: I think that lvgl must take into account dedicated parameters for screens. it will help to make the library evolve towards modern features but it can be difficult.
e.g organization:
```
 -lvgl
     -screen 1 handle
              - gpio handle
              - spi handle
     -screen 2 handle
              - par handle
```
Like you said, use VDB chunk send to the good screen when finish to virtual draw. I wait your feedback about how complex can be the changes

### DEVICE HANDLER
Maybe we need take a look on [Device Tree](https://github.com/devicetree-org) organization but is advanced. Some clue for abstraction can help.
To my point lvgl need for multi-display the screen handle. e.g., for ssd1306, it will be `ssd1306_t* handle`. I do not see how to use enumerator can simplify the thing.
each screen descriptor got  bus device descriptor handle (I2C/SPI/GPIO...)
That means the user can customize the bus device descriptor  (see below).

- Result: A good way to resolve with this new lvgl API form (if you do it) is to add an define option on lvgl. To enable external driver support. (it will disable multi-display features  and use legacy flush/map/fill ) or people use lv_drivers with this lvgl advanced features enable (multi_display, reliability, ...).

### FUNCTION NAMES
For function names, was just a first approach to see what is needed. I would like to reach a perfect abstraction of these peripherals. ( let's try )

**delay** : Nothing to say ^^, was simple.
**gpio**: In every SDK i used a unique identifier per pin. I thought that would be enough.
Anyway i  can replace it by a "gpio_dev". The driver will not know the port directly but just pass it directly the information set by the user.
e.g. :
```c
typedef struct {
    int gpio;
    int port;
} gpio_dev_t ;
lv_gpio_write(void* gpio_dev, uint8_t val);
```
There could therefore be several different ports and several different pins.

**i2c**: the i2c_dev handle we got the adrees (like GPIO). e.g.:
```c
typedef struct {
    uint8_t bus;
    uint8_t addr;
    (others configs);
} i2c_dev_t ;
```
I prefer use this kind of programmation because the user can use parameters specific to his SDK / drivers configurations. I am glad to see speak about 8/16 bits register. I did a job as you describe it.I had already discussed this problem in the esp-open-rtos project. In fact, 8-bit register components are used more often than 16-bit register components. So the user should add the i2c function 16-bit register only if needed by a specific driver. We should add comments on the dependencies to know which API is used by which driver as in lvgl on the screen objects. In practical, that mean the second API is used rarely.
e.g:
`/*SSD1306 (dependencies: i2c 8 bit / spi  / gpio )*/`

**Spi**:
I will take a second look about it and try with the current driver. I ask myself if we can add a define like
 e.g. "HW_SPI_BOOST" with special function but optional.

- Result: I think we need more feedback on API to improve it. I will take into account your comment and changes some things, then i think you can create a experimental branch for this pull request. And wait feedback and more drivers.

### Other
Thanks for the SSD1306 driver comment. To be honest i didn't start it. It is taken from
esp-open-rtos sdk. You can see people who work on it (like me). My work on this driver can be resume
by: I2c improvements, SPI3 protocol added, SH1106 compatibility added, HW Scrolling added, fixes bug, portability...)
I will add author and maybe copyright flag to the file.  (It is BSD or MIT, i will check)
I will design ILI9341 driver in same way ( protocols / command / .... )

kisvegabor:
Hi,

I created a branch: **new_api** and I will merge your changes there.

### MULTI DISPLAY SUPPORT

I see two different approach of using multiple screens:
1. Extend the screen size: like two monitors and you can drag a window between screen. It need a driver level support to send the right area to the right display. Easy to do.
2. Independent displays: e.g. a main TFT, a small monochrome, any VNC. LittlevGL need to handle them independently. I will do it, because it's quite difficult and need very deep understanding of the library.

How does it sounds to you?

### DEVICE HANDLER

I like the approach of creating handlers for "everything". It makes the system flexible.

For enums I mean:
```c
typedef enum {
  LV_I2C_0,
  LV_I2C_1,
  LV_I2C_2,
  LV_I2C_3,
  LV_I2C_4,
}lv_i2c_bus_t;

typedef struct {  //From your example
    lv_i2c_bus_t bus;
    uint8_t addr;
    (others configs);
} i2c_dev_t ;
```
We should design device handlers very extensively.

So I will merge your changes, to see it in the practice!

Zaltora:
Nice,

The second option sound good :) (my favorite).  The first is good if screen are the same.
for device handlers, you want create a struct by default ?. I was thinking use the SDK bus handles ( that why i use void* ). It will help user to use it i guess ? He can use the build-in handler or just use a typedef to use the SDK bus handler. Use enum make sense with build-in handle struct.

kisvegabor:
Multi display: Great! :)

Device handler: my only concern about using pointer that not all SDK (driver library) handles drivers with pointers. For example if you write your drivers yourself you just configure registers by hand and don't have a complex system of device handlers. That is why I vote for enums because they are compatible with every implementation. You just need to add a switch-case to the driver. What do you think?

kisvegabor:
Excellent idea! Then the enum is not required at all because user can use simple numbers too, or define his own enum if required. So it become up to the user how to select the periphery.

PR #3, opened 2018-02-05
API improvements / ILI9341

Zaltora:
Hi,
Since out last discussion, i added the handle typedef. I will rework SPI after. We got a lot of possibility now with the interface API.

I begin to write the ILI9341 driver. We can found many driver on internet but driver is only for SPI, and a lot of command are not described, just hex datasend. ILI9341 datasheet got some errors too. I will details all command i write. The register config of ili9341 are pretty chaotic so i use bit field struct to set value. and i copy the structure in to an array predefined with fixed value (see datasheet). that means user can't change unknown bits. 

Main features planned for this lib:
-All commands integration (lvl 0 API)
-More protocol supported (8080, SPI3, ... )

No need to merge now. I open this thread to discuss about features and advancements.

### DIFF
--- a/display/ILI9341.c
+++ b/display/ILI9341.c
@@ -1,0 +1,646 @@
+/**
+ * @file ILI9341.c
+ * @author zaltora  (https://github.com/Zaltora)
+ * @copyright MIT License.
+ */
+//TODO: remove this
+#define USE_ILI9341 1
+/*********************
+ *      INCLUDES
+ *********************/
+#include "ILI9341.h"
+#if USE_ILI9341 != 0
+
+#include <stdbool.h>
+#include "lvgl/lv_core/lv_vdb.h"
+
+
+//TODO: remove this
+#define ILI9341_DEBUG 1
+
+/*********************
+ *      DEFINES
+ *********************/
+#ifndef ILI9341_SPI4_SUPPORT
+#define ILI9341_SPI4_SUPPORT 1
+#endif
+
+#ifndef ILI9341_SPI3_SUPPORT
+#define ILI9341_SPI3_SUPPORT 1
+#endif
+
+#ifndef ILI9341_MANUAL_DC
+#define ILI9341_MANUAL_DC 1
+#endif
+
+#ifndef ILI9341_MANUAL_CS
+#define ILI9341_MANUAL_CS 1
+#endif
+
+/* utility */
+#define BIT_MASK(a, b) (((unsigned) -1 >> (31 - (b))) & ~((1U << (a)) - 1))
+
+/* ILI9341 regular commands */
+#define ILI9341_NO_OPERATION            (0x00)
+#define ILI9341_SOFT_RESET              (0x01)
+#define ILI9341_DIS_INFOS               (0x04)
+#define ILI9341_DIS_STATUS              (0x09)
+#define ILI9341_DIS_POWER_MODE          (0x0A)
+#define ILI9341_DIS_MADCTL              (0x0B)
+#define ILI9341_DIS_PIXEL_FMT           (0x0C)
+#define ILI9341_DIS_IMAGE_FMT           (0x0D)
+#define ILI9341_DIS_SIGNAL_MODE         (0x0E)
+#define ILI9341_DIS_SELF_DIA_RESULT     (0x0F)
+#define ILI9341_SLEEP_ON                (0x10)
+#define ILI9341_SLEEP_OFF               (0x11)
+#define ILI9341_PARTIAL_MODE            (0x12)
+#define ILI9341_NORMAL_MODE             (0x13)
+#define ILI9341_INVERSION_OFF           (0x20)
+#define ILI9341_INVERSION_ON            (0x21)
+#define ILI9341_GAMMA_SET               (0x26)
+#define ILI9341_DIS_OFF                 (0x28)
+#define ILI9341_DIS_ON                  (0x29)
+#define ILI9341_COLUMN_ADDR_SET         (0x2A)
+#define ILI9341_PAGE_ADDR_SET           (0x2B)
+
+#define ILI9341_MEMORY_WRITE            (0x2C)
+#define ILI9341_COLOR_SET               (0x2D)
+#define ILI9341_MEMORY_READ             (0x2E)
+#define ILI9341_PARTIAL_AREA            (0x30)
+#define ILI9341_VERT_SCROLLING          (0x33)
+#define ILI9341_TEARING_LINE_OFF        (0x34)
+#define ILI9341_TEARING_LINE_ON         (0x35)
+#define ILI9341_MEMORY_ACCES_CTR        (0x36)
+#define ILI9341_VERT_SCROLL_START_ADDR  (0x37)
+#define ILI9341_IDLE_MODE_OFF           (0x38)
+#define ILI9341_IDLE_MODE_ON            (0x39)
+#define ILI9341_PIXEL_FMT_SET           (0x3A)
+#define ILI9341_WRITE_MEMORY_CONTINUE   (0x3C)
+#define ILI9341_READ_MEMORY_CONTINUE    (0x3E)
+#define ILI9341_SET_TEAR_SCANLINE       (0x44)
+#define ILI9341_GET_SCANLINE            (0x45)
+#define ILI9341_DIS_SET_BRIGHTNESS      (0x51)
+
+#define ILI9341_DIS_GET_BRIGHTNESS      (0x52)
+#define ILI9341_WRITE_CTRL_DIS          (0x53)
+#define ILI9341_READ_CTRL_DIS           (0x54)
+#define ILI9341_WRITE_ADAP_BRIGTH_CTR   (0x55)
+#define ILI9341_READ_ADAP_BRIGTH_CTR    (0x56)
+#define ILI9341_WRITE_CABC_MIN_BRIGTH   (0x5E)
+#define ILI9341_READ_CABC_MIN_BRIGTH    (0x5F)
+#define ILI9341_READ_ID1                (0xDA)
+#define ILI9341_READ_ID2                (0xDB)
+#define ILI9341_READ_ID3                (0xDC)
+
+/* ILI9341 Extended commands */
+
+#define ILI9341_RGB_INT_SIG_CTR         (0xB0)
+#define ILI9341_FRAME_CTR_NORMAL        (0xB1)
+#define ILI9341_FRAME_CTR_IDLE          (0xB2)
+#define ILI9341_FRAME_CTR_PARTIAL       (0xB3)
+#define ILI9341_DIS_INVERSION_CTR       (0xB4)
+#define ILI9341_BLANK_PORCH_CTR         (0xB5)
+#define ILI9341_DIS_FUNCTION_CTR        (0xB6)
+#define ILI9341_ENTRY_MODE_SET          (0xB7)
+#define ILI9341_BACKLIGTH_CTR_1         (0xB8)
+#define ILI9341_BACKLIGTH_CTR_2         (0xB9)
+#define ILI9341_BACKLIGTH_CTR_3         (0xBA)
+#define ILI9341_BACKLIGTH_CTR_4         (0xBB)
+#define ILI9341_BACKLIGTH_CTR_5         (0xBC)
+#define ILI9341_BACKLIGTH_CTR_7         (0xBE)
+#define ILI9341_BACKLIGTH_CTR_8         (0xBF)
+#define ILI9341_PWR_CTR_1               (0xC0)
+#define ILI9341_PWR_CTR_2               (0xC1)
+#define ILI9341_VCOM_CTR_1              (0xC5)
+#define ILI9341_VCOM_CTR_2              (0xC7)
+#define ILI9341_NV_MEMORY_WRITE         (0xD0)
+#define ILI9341_NV_MEMORY_PROTECT_KEY   (0xD1)
+#define ILI9341_NV_MEMORY_STATUS        (0xD2)
+#define ILI9341_READ_ID4                (0xD3)
+#define ILI9341_POS_GAMMA_COR           (0xE0)
+#define ILI9341_NEG_GAMMA_COR           (0xE1)
+#define ILI9341_DIG_GAMMA_CTR1          (0xE2)
+#define ILI9341_DIG_GAMMA_CTR2          (0xE3)
+#define ILI9341_UNKNOW                  (0xEF)
+#define ILI9341_INTERFACE_CTR           (0xF6)
+
+#define ILI9341_PWR_CTR_A               (0xCB)
+#define ILI9341_PWR_CTR_B               (0xCF)
+#define ILI9341_TIMING_CTR_A            (0xE8)
+#define ILI9341_UNKNOW_2                (0xE9)  //This cmd exist ?
+#define ILI9341_TIMING_CTR_B            (0xEA)
+#define ILI9341_PWR_ON_SEQ_CTR          (0xED)
+#define ILI9341_ENABLE_3G               (0xF2)
+#define ILI9341_PUMP_RATIO_CTR          (0xF7)
+
+
+/**********************
+ *      TYPEDEFS
+ **********************/
+
+/**********************
+ *  STATIC PROTOTYPES
+ **********************/
+static int _sendCommand(uint8_t cmd);
+static int _sendCommandData(uint8_t cmd, uint8_t* data, uint8_t len);
+
+/**********************
+ *  STATIC VARIABLES
+ **********************/
+
+
+/**********************
+ *      MACROS
+ **********************/
+#if ILI9341_DEBUG
+#include <stdio.h>
+#define debug(fmt, ...) printf("%s: " fmt "\n", "ILI9341", ## __VA_ARGS__)
+#else
+#define debug(fmt, ...)
+#endif
+
+/**********************
+ *   GLOBAL FUNCTIONS
+ **********************/
+
+void ili9341_flush(int32_t x1, int32_t y1, int32_t x2, int32_t y2, const lv_color_t * color_p)
+{
+    /*Return if the area is out the screen*/
+    if (x2 < 0 || y2 < 0 || x1 > LV_HOR_RES - 1 || y1 > LV_VER_RES - 1)
+    {
+        lv_flush_ready();
+        return;
+    }
+
+    /*Truncate the area to the screen*/
+    int32_t act_x1 = x1 < 0 ? 0 : x1;
+    int32_t act_y1 = y1 < 0 ? 0 : y1;
+    int32_t act_x2 = x2 > LV_HOR_RES - 1 ? LV_HOR_RES - 1 : x2;
+    int32_t act_y2 = y2 > LV_VER_RES - 1 ? LV_VER_RES - 1 : y2;
+
+
+    lv_flush_ready();
+}
+
+void ili9341_fill(int32_t x1, int32_t y1, int32_t x2, int32_t y2, lv_color_t color)
+{
+    /*The most simple case (but also the slowest) to put all pixels to the screen one-by-one*/
+
+    /*Return if the area is out the screen*/
+    if (x2 < 0 || y2 < 0 || x1 > LV_HOR_RES - 1 || y1 > LV_VER_RES - 1)
+    {
+        return;
+    }
+
+    /*Truncate the area to the screen*/
+    int32_t act_x1 = x1 < 0 ? 0 : x1;
+    int32_t act_y1 = y1 < 0 ? 0 : y1;
+    int32_t act_x2 = x2 > LV_HOR_RES - 1 ? LV_HOR_RES - 1 : x2;
+    int32_t act_y2 = y2 > LV_VER_RES - 1 ? LV_VER_RES - 1 : y2;
+
+}
+
+void ili9341_map(int32_t x1, int32_t y1, int32_t x2, int32_t y2, const lv_color_t * color_p)
+{
+    /*Return if the area is out the screen*/
+    if (x2 < 0 || y2 < 0 || x1 > LV_HOR_RES - 1 || y1 > LV_VER_RES - 1)
+    {
+        return;
+    }
+
+    /*Truncate the area to the screen*/
+    int32_t act_x1 = x1 < 0 ? 0 : x1;
+    int32_t act_y1 = y1 < 0 ? 0 : y1;
+    int32_t act_x2 = x2 > LV_HOR_RES - 1 ? LV_HOR_RES - 1 : x2;
+    int32_t act_y2 = y2 > LV_VER_RES - 1 ? LV_VER_RES - 1 : y2;
+
+}
+
+/* Perform default init routine according */
+int ili9341_init(const ili9341_t *dev)
+{
+    int err;
+    switch (dev->height)
+    {
+    case 240:
+        break;
+    case 320:
+        break;
+    default:
+        debug("Unsupported screen height");
+        return -ENOTSUP;
+    }
+
+    switch (dev->protocol)
+    {
+#if (ILI9341_SPI4_SUPPORT)
+    case ILI9341_PROTO_SERIAL_8BIT:
+        break;
+#endif
+#if (ILI9341_SPI3_SUPPORT)
+    case ILI9341_PROTO_SERIAL_9BIT:
+        break;
+#endif
+    default:
+        debug("Unsupported protocol");
+        return -EPROTONOSUPPORT;
+    }
+
+
+    if((err = ili9341_unknow(dev)))
+        return err;
+
+    ili9341_pwr_ctrl_b_t pwr_ctrl_b_config = { 0 };
+    pwr_ctrl_b_config.dc_ena = 1 ;
+    pwr_ctrl_b_config.pceq = 1 ;
+
+    if((err = ili9341_power_control_b(dev, pwr_ctrl_b_config)))
+        return err;
+
+    ili9341_pwr_seq_ctrl_t pwr_seq_ctrl_config  = { 0 };
+    pwr_seq_ctrl_config.cp23_soft_start = 0 ;
+    pwr_seq_ctrl_config.cp1_soft_start = 2 ;
+    pwr_seq_ctrl_config.en_ddvdh = 3 ;
+    pwr_seq_ctrl_config.en_vgh = 1 ;
+    pwr_seq_ctrl_config.en_vgl = 2 ;
+    pwr_seq_ctrl_config.ddvdh_enh = 1;
+
+    if((err = ili9341_power_on_seq_ctrl(dev, pwr_seq_ctrl_config)))
+        return err;
+
+    ili9341_timing_ctrl_a_t timing_ctrl_a_config = { 0 };
+    timing_ctrl_a_config.now = 1 ;
+
+    if((err = ili9341_timing_ctrl_a(dev, timing_ctrl_a_config)))
+        return err;
+
+    ili9341_pwr_ctrl_a_t power_ctrl_a_config = { 0 };
+    power_ctrl_a_config.reg_vd = 4 ;
+    power_ctrl_a_config.vbc = 2 ;
+
+    if((err = ili9341_pwr_ctrl_a(dev, power_ctrl_a_config)))
+        return err;
+
+    ili9341_pump_ratio_ctrl_t pump_ration_ctrl_config = { 0 };
+    pump_ration_ctrl_config.ratio = 2 ;
+
+    if((err = ili9341_pump_ratio_ctrl(dev, pump_ration_ctrl_config)))
+        return err;
+
+    ili9341_timing_ctrl_b_t timing_ctrl_b_config = { 0 };
+    if((err = ili9341_timing_ctrl_b(dev, timing_ctrl_b_config)))
+        return err;
+
+    ili9341_pwr_ctrl_1_t pwr_ctrl_1_config = { 0 } ;
+    pwr_ctrl_1_config.vrh = 0b100011;
+    if((err = ili9341_pwr_ctrl_1(dev, pwr_ctrl_1_config)))
+        return err;
+
+    ili9341_pwr_ctrl_2_t pwr_ctrl_2_config = { 0 } ;
+    if((err = ili9341_pwr_ctrl_2(dev, pwr_ctrl_2_config)))
+        return err;
+
+    ili9341_vcom_ctrl_1_t vcom_ctrl_1_config = { 0 } ;
+    vcom_ctrl_1_config.vmh = 0b00111110;
+    vcom_ctrl_1_config.vml = 0b00101000;
+    if((err = ili9341_vcom_ctrl_1(dev, vcom_ctrl_1_config)))
+        return err;
+
+    ili9341_vcom_ctrl_2_t vcom_ctrl_2_config = { 0 } ;
+    vcom_ctrl_2_config.vmf = 6 ;
+    vcom_ctrl_2_config.nvm = 1 ;
+    if((err = ili9341_vcom_ctrl_2(dev, vcom_ctrl_2_config)))
+        return err;
+
+    ili9341_mem_ctrl_t ili9341_mem_ctrl_config = { 0 } ;
+    ili9341_mem_ctrl_config.mx = 1 ;
+    ili9341_mem_ctrl_config.bgr = 1 ;
+    if((err = ili9341_mem_ctrl(dev, ili9341_mem_ctrl_config)))
+        return err;
+
+    ili9341_vert_scroll_start_t ili9341_vert_scroll_start_config = { 0 };
+    if((err = ili9341_vert_scroll_start(dev, ili9341_vert_scroll_start_config)))
+        return err;
+
+    ili9341_px_fmt_t ili9341_pixel_fmt_config = { 0 };
+    ili9341_pixel_fmt_config.dbi = 0b101 ;
+    ili9341_pixel_fmt_config.dpi = 0b101 ;
+    if((err = ili9341_pixel_fmt(dev, ili9341_pixel_fmt_config)))
+        return err;
+
+    ili9341_frame_rate_ctrl_t ili9341_frame_rate_ctrl_config = { 0 };
+    ili9341_frame_rate_ctrl_config.rtna = 0b11000 ;
+    if((err = ili9341_frame_rate_ctrl(dev, ili9341_frame_rate_ctrl_config)))
+        return err;
+
+    ili9341_dis_fn_ctrl_t ili9341_dis_fn_ctrl_config = { 0 };
+    ili9341_dis_fn_ctrl_config.ptg = 0b10 ;
+    ili9341_dis_fn_ctrl_config.isc = 0b0010 ;
+    ili9341_dis_fn_ctrl_config.rev = 0b1 ;
+    ili9341_dis_fn_ctrl_config.nl = 0b100111;
+    if((err = ili9341_display_fn_ctrl(dev, ili9341_dis_fn_ctrl_config)))
+        return err;
+
+    ili9341_ena_3g_t ili9341_ena_3g_config = { 0 };
+    if((err = ili9341_enable_3g(dev, ili9341_ena_3g_config)))
+        return err;
+
+    ili9341_gamma_set_t ili9341_gamma_set_config = { 0 };
+    ili9341_gamma_set_config.gamma_set = 0x01 ;
+    if((err = ili9341_gamma_set(dev, ili9341_gamma_set_config)))
+        return err;
+
+
+    ili9341_gamma_cor_t ili9341_gamma_config = { 0 };
+    ili9341_gamma_config.v63 = 0x0F ;
+    ili9341_gamma_config.v62 = 0x31 ;
+    ili9341_gamma_config.v61 = 0x2B ;
+    ili9341_gamma_config.v59 = 0x0C ;
+    ili9341_gamma_config.v57 = 0x0E ;
+    ili9341_gamma_config.v50 = 0x08 ;
+    ili9341_gamma_config.v43 = 0x4E ;
+    ili9341_gamma_config.v36 = 0x0F ;
+    ili9341_gamma_config.v27 = 0x01 ;
+    ili9341_gamma_config.v20 = 0x37 ;
+    ili9341_gamma_config.v13 = 0x07 ;
+    ili9341_gamma_config.v6 = 0x10 ;
+    ili9341_gamma_config.v4 = 0x03 ;
+    ili9341_gamma_config.v2 = 0x0E ;
+    ili9341_gamma_config.v1 = 0x09 ;
+    ili9341_gamma_config.v0 = 0x00 ;
+    if((err = ili9341_gamma_cor(dev, ili9341_gamma_pos, ili9341_gamma_config)))
+        return err;
+
+    ili9341_gamma_config.v63 = 0x00 ;
+    ili9341_gamma_config.v62 = 0x0E ;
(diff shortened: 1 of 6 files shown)


kisvegabor:
Hi,

great hank you! Well configurable drivers would be very good. They are missing now...

I was thinking about the new API and the device handler and got the following (IMHO very clean) structure:
1. Extend lv_disp_drv_t with a `void * dev;` field.
2. User can put here any structure he want, like: SPI dev + IO etc (pointer to a user defined structure)
3. The display/touch pad driver just pass the `dev` without any knowledge about it. 
4. The MCU drivers (which are written by the user) get the dev and user can "parse" it as he want. E.g. we could have  a function like these: `lv_drv_spi_cs_low(dev);`

Speaking about SPI I can imagine functions like these:
```c
lv_drv_spi_wr_cs(dev, 0 or 1);
lv_drv_spi_wr_dc(dev, 0 or 1);
lv_drv_spi_wr_byte(dev, data);
lv_drv_spi_wr_array(dev, array);
```

I the user want to use HW controller CS then he can leave `lv_drv_spi_wr_cs` empty and do it in  `lv_drv_spi_wr_byte`.

For me it is very clear, flexible and easy to integrate for the user too. What do you think?

Zaltora:
I will update SPI as you said because it is a great idea. I was doing this before and i stopped i don't no why ? ( I speak about dc and cs pin API)
So now i confirm, bus will manage external pin too. It is work too with the parallel protocol. I will take a look what is the best way to do with the ili9341 datasheet. Maybe add "soft driver" for this protocol (8080) here too. that will help users because i don't found driver on it. If you do, share ;)

I remain perplexed for:
```c
lv_drv_spi_wr_byte(dev, data);
lv_drv_spi_wr_array(dev, array, len);
```
 Spi is a full-duplex protocol, it is make more sense to get function on it like:
(maybe it is already the case, but you miss return so ^^ )
```c
uint8_t lv_drv_spi_wr_byte(dev, data);
void lv_drv_spi_wr_array(dev, array_in, array_out, len);
```
Anyway when the array is NULL, that mean the transaction will bi unidirectional. Len will be bit sized, not byte sized. Because we got protocol that use 9 bit transaction with mostly screen. And it is more flexible if user don't use advanced features from spi.
Now second question, why add a function specific to byte ? i guess we can do it like for uin16_t and uint32_t  but maybe it is does a lot of config ? why not use the generic array send for all.

A useful function on SPI exist too, it is the repeat function. We can add an api to this too.

My problem was to add the advanced features from spi like address, cmd, dummy,..
So i will add a define for enable theses features because it is permit to accelerate spi transaction in case like SPI3 protocol and device that begin always with same thing during transaction.
`#define SPI_HW_ADV`

So it is a template of the final API if i use all possibility:
```c
//pin control
void lv_drv_spi_wr_cs(dev, 0 or 1);
void lv_drv_spi_wr_dc(dev, 0 or 1);
//popular type transaction
uint8_t lv_drv_spi_wr8(dev, data);
uint16_t lv_drv_spi_wr16(dev, data);
uint32_t lv_drv_spi_wr32(dev, data);
//custom sent/receive
void lv_drv_spi_wr_array(dev, array_in, array_out, len*8);
//repeat API for popular type
void lv_drv_spi_repeat8(dev, data, len);
void lv_drv_spi_repeat16(dev, data, len);
void lv_drv_spi_repeat32(dev, data, len);
//spi advances features
#if SPI_HW_ADV
void lv_drv_spi_address(dev, address);
void lv_drv_spi_command(dev, address);
void lv_drv_spi_dummy(dev, address);
void lv_drv_spi_clear_address(dev);
void lv_drv_spi_clear_command(dev);
void lv_drv_spi_clear_dummy(dev);
#endif
```
IMHO, i guess you got all common spi function possible above. I need think more about it with some feedback.
I ask myself if we add function for each type or just add generic one (for repeat for example). that means 2 functions instead of 6.

I confirm the fact that the generic device_handler will handle pin and more. Finally the driver will just use the protocol API to do this. what feeling about the SPI api ?

kisvegabor:
So finally do you agree with the user defined `dev`?

True, the SPI is bidirectional and the functions should support it. I also like to define the length in bits instead of bytes.
 
About lv_drv_spi_wr8/16/32: why not use the general wr_array for these purpose too?

About repeat functions: they are useful when the you fill with a color, right? In the next major release (v6.0) I'm planning to  remove the fill and map functions, so the operation without VDB because I can't see any advantage of it. So the repeat function won't be required then. 

About SPI_HW_ADV: Can you explain how to use these functions? What does these functions do?

A little off topic and explanation why I try to join and remove functions:  
Once I heard a proverb: something is not perfect when you can't add anything, but when you can't take away anything. Since then I try to follow this :) Try to keep everything as simple as possible.

Zaltora:
> So finally do you agree with the user defined dev?

Maybe we get misunderstood. I thought that the proposed idea was identical to the original version of the lib and got a cs/dc write in hard. ^^

> True, the SPI is bidirectional and the functions should support it. I also like to define the length in bits instead of bytes.

Another point for futures SPI usage. We need to think about Dual/quad SPI too. Some devices use it, 
[FT811](http://www.ftdichip.com/Products/ICs/FT81X.html)

> About lv_drv_spi_wr8/16/32: why not use the general wr_array for these purpose too?

this is why i said i can do 2 functions instead of 6. If we use 8/16/32 call, it will be faster than use a generic one ( or not depending of sdk). I vote for generic one. Anyway, it is "alpha" version API. Optimization come later with feedback.

> About SPI_HW_ADV: Can you explain how to use these functions? What does these functions do?

This is an interesting feature. Before begin I will concatenate them in 2 functions.
```c
void lv_drv_spi_wr_fix(dev, type, data);
void lv_drv_spi_clear_fix(dev, type);
```
Theses function will set a fixed value for each future transaction in this device. If you got device that begin with always same thing like a address:
 <------ADDR------> <------DATA0------> .....  <------ADDR------> <------DATAn------> .....
or a d/c bit value:
<-DC-><-D7-><-D6-><-D5-><-D4-><-D3-><-D2-><-D1-><-D0->....

Theses functions are usefull on DUAL/QUAD SPI device too . Because you need  set dummy bits before writing and reading (or addr or cmd or all !).
When you use the command SPI instead of just basi cwrite, the HW  know already it will start to read MISO line after this addr/cmd/dummy configs
Here a link: [SPI infos](https://www.corelis.com/education/SPI_Tutorial.htm)

Final API  (WIP template):
```c
//pin control
void lv_drv_spi_wr_cs(dev, 0 or 1);
void lv_drv_spi_wr_dc(dev, 0 or 1);
//transfert
void lv_drv_spi_wr_array(dev, array_in, array_out, len*8);
void lv_drv_spi_repeat(dev, data, wsize, len);  //wsize: word size: 1,2, 4 bytes
//spi advances features
void lv_drv_spi_wr_fix(dev, type, data); // type: addres, commad, dummy
void lv_drv_spi_clear_fix(dev, type);
```
Finally, theses functions are useful to set extra bits without need configure each time this line. The user can implement himself this functionality too (if not implemented in the sdk for example).
e.g
```c
typedef struct {
   uint8_t bus;
   uint8_t cs;
   uint8_t dc;
   uint8_t addr;
   uint8_t dummy;
   uint8_t cmd;
   uint8_t flags_extra_bit;
};
```

Zaltora:
> About repeat functions: they are useful when the you fill with a color, right? In the next major release (v6.0) I'm planning to remove the fill and map functions, so the operation without VDB because I can't see any advantage of it. So the repeat function won't be required then.

Yes, used for fill color or erase the screen RAM entirely.  Maybe repeated fn not needed like you said. Need finish driver and do test.

If you rework API, can i suggest something. Fill color is very useful and can be used to draw "area" fast. Some screen can fill an area with a color just by giving x1,y1,x2,y2 and c without need send each pixel. it is feasible that the VDB can choice "fill draw" when he need  set a color on a area before draw next layer ?

kisvegabor:
I suggest to enumerate the points :)

1. User defined dev:  I'm afraid I don't understand you. :'( What identical with what? 

2. Dual and QSPI support: Yes, you are right. It seems will be more common in the future too! So we should be sure that the API supports them well.

3.  lv_drv_spi_wr8/16/32: Great, I also vote for the generic :)

4. repeat and fill: Yes, the GPU in external display controllers are the only reason for `disp_fill()`. Now the real draw function (which draws not to VDB but to the frame buffer directly) are always in the code. Then at least they need to be disabled when LV_VDB_SIZE != 0 and keep the disp_fill option for LV_VDB_SIZE == 0

Zaltora:
1 . identical with current system macro. one macro for display, one macro for indev. not enough  flexible. i have misunderstood anyway, so i close it !! ^^ Now it is clear to me ;)
2.  Ok :) I don't think we need add additionnal function. D/Q spi are initialized by user i guess.
3.  Ok :)
4.  Ok then. don't know well mechanics to optimize this.

I change my mind about data transfer in bits since anyway, special case are for screen with 9bit data transfert for example and can be done with `lv_drv_spi_wr_fix`. i don't imagine user manage a 9 bit array. It is difficult and need additional treatments.
But instead i want to add word_size like for standard stdio API.

Final API (WIP template):
```c
//pin control
void lv_drv_spi_wr_cs(dev, 0 or 1);
void lv_drv_spi_wr_dc(dev, 0 or 1);
//transfert
void lv_drv_spi_wr_array(dev, array_in, array_out, wsize, len);  //len in bytes
void lv_drv_spi_repeat(dev, data, wsize, len);  //wsize: word size: 1,2, 4 bytes
//spi advances features
void lv_drv_spi_wr_fix(dev, type, data); // type: addres, commad, dummy
void lv_drv_spi_clear_fix(dev, type);
```

kisvegabor:
1. Okay :)
2. Yes, Simple or D/Q SPI seems similar from our side. I
3. -  
4. One think came to my mind about repeat: if  `disp_fill()`'s only purpose is to ask the external display controller to fill an area (if your display controller don't support it you should use VDB) then repeat is become unnecessary again. Do you agree?
5. 9bit: you are right quite difficult to manage it in the practice. But I can't imagine how does it work with `lv_drv_spi_wr_fix`. For example when you need to send a Command/Data bit first?

Thank you for your persistent work! :)
I think if we well establish the SPI interface the others will go much faster because will have same mindset.

Zaltora:
4. I agree but I suggest to keep the function. After all, if it's not used, we'll delete it later. 
5. I will show you an example like i use in SSD1306 and maybe in ILI9341 driver

```c
//This command will set 1 bit Command for each transaction
//The bit will be D/C and this level is 0. Data Mode
//You can use addr or cmd register to this. Not important here.
lv_drv_spi_set(spi_dev, eCommand, 0, 1);

//We use a loop here instead of this: lv_drv_spi_tr_array(spi_dev, NULL, buf, 1, len);
//The reason is simple, the screen need receive D/C bit for each byte (for SSD1306 protocol)
//When you directly use the function, the frame look like this:
// <-DC-><-BYTE0-><-BYTE1-><-BYTEx->...<-DC-><-BYTEx+1->...<-BYTEn-> 
// Depend of the SPI FIFO size. Using a loop ensure SPI to close and reopen a transaction for one byte.
// <-DC-><-BYTE0-><-DC-><-BYTE1-><-DC-><-BYTEn->
for (i = 0; i < len; i++)
{
     //Send byte one by one with a D/C bit value configured before and static.
     //<-DC-><-buf[i]->
     lv_drv_spi_tr_array(spi_dev, NULL, &buf[i], 1, 1); //1 byte array
}
//Remove the D/C bit.
 lv_drv_spi_clear(spi_dev, eCommand);
```

Final API (Template)
```c
//pin control
void lv_drv_spi_wr_cs(dev, 0 or 1);
void lv_drv_spi_wr_dc(dev, 0 or 1);
//transfert
void lv_drv_spi_tr_array(dev, array_in, array_out, wsize, len);  //len in bytes
void lv_drv_spi_wr_repeat(dev, data, wsize, len);  //wsize: word size: 1,2, 4 bytes
//spi advances features
void lv_drv_spi_set(dev, type, data, blen); // type: addres, commad, dummy ; //blen, bit length
void lv_drv_spi_clear(dev, type);
```

In software mode if not hw setting present:
first function will set a variable `uint8_t command =  0;  uint8_t  bit_length = 1 ;`
when you do transaction you can setup like this before transfer. e.g:
```c
if(flag_cmd)
{
uint16_t data =  (command << 8 & 0xF0 ) | data ;
}
```
final user need to setup spi to send 1 bit more. If it is impossible to do it for him, he don't use the protocol SPI3. ( Can be disable in configuration)

I know a lot SPI problem, playing a lot with registers in esp8266 components and oscilloscope to view result. :). 
the next stage (when i will do ILI9341 driver) will be to write a  software 8080/6080 driver with gpio control ^^.
Do an abstraction like this is very cool because he can be used for other purpose. I will use it to make my app portable between multiple component/sdk. with only one file abstraction API to change.

kisvegabor:
Hi,

4. Okay, keep the repeat and we will see. 

5. Thank you for the detailed explanation! I got the idea now. But as I see user can do the same thing in the `lv_drv_spi_wr_dc()` function too (Save the mode and reconfigure the SPI). Is there any other purpose of using  `lv_drv_spi_set()` except changing between Data and Command mode?

Zaltora:
HI,

5. You are right but you forget a few cases:
- If you use `lv_drv_spi_wr_dc() `, you can set the bit to 1 or 0 but not delete the bit. If the SPI from SDK is difficult to manage with multiple device, maybe it is interesting to remove extra bit from a custom transaction.
- `lv_drv_spi_set()` can be used with drivers that use DIO/QIO spi drivers and need set dummy, addr or cmd bits.
- If we use DC bit with `lv_drv_spi_wr_dc() `, we will need extra check code when a transaction begin. For example, a device need 8 bit transaction. Before send the data. it will check if the extra bit is set and delete it if yes. then it will check for each transaction if the bit is set.

kisvegabor:
Hi,

I see. Only one things left in my mind: `lv_drv_spi_set` is not describes what the function does. What about `lv_drv_spi_set/clr_preemble` ?

Zaltora:
Changed API. 
Tested with esp8266 and SSD1306. It is work well. Custom typedef help to port the code.

According to new information and with multiples warning with gcc, i change again "inline system". It is a hard subject. More information about it here : [extern inline](https://blogs.oracle.com/d/inline-functions-in-c).
It seem that some compilers got different behavior to hold this system. That no help for portability. I use static inline instead and remove the *.c file. 
I discovered that gcc got a attribute `__attribute__((always_inline))` that can be added with a MACRO for people that want it.

kisvegabor:
Hi,
It looks very well! Congratulation!  

Meanwhile I ported LittlevGL to ESP32 with ILI9341 and now I understand your goals better. I got familiar with the SPI handling in ESP32 and saw the address and command bit options. 

A few things are not clear to me:
- Now user needs to initialize an ssd1306_t variable an pas it to the driver somehow? I don't understand where the 'dev' variable came form.
- I can't see `_flag_redraw` where is handled. (Where happens the drawing?)

Zaltora:
Maybe i will answer you with my code. here it is:

in lv_drv_conf.h:
```c
typedef const void* lv_par_handle_t ;

typedef struct {
    uint8_t bus;
    uint8_t cs;
    uint8_t dc;
} lv_spi_dev_t, *lv_spi_handle_t;

typedef const i2c_dev_t *lv_i2c_handle_t ;
typedef uint8_t lv_gpio_handle_t ;
```
in my screen file:
```c
static uint8_t aucBuffer[DISPLAY_WIDTH * DISPLAY_HEIGHT / 8]; //Screen buffer

/* Declare spi descriptor */
static lv_spi_dev_t _xSpiDev =
{
    .bus = SPI_BUS,
    .cs = PIN_IHM_CS,
    .dc = LV_DRIVER_NOPIN,
};

/* Declare device descriptor */
static ssd1306_t xScreenDev;

void vScreenInitialization()
{
    //Bus initialization
    //portable_gpio_enable(PIN_IHM_CS, PORT_GPIO_MODE_OUTPUT);
    //portable_gpio_write(PIN_IHM_CS, true);
    spi_init(SPI_BUS, SPI_MODE0, SPI_FREQ_DIV_8M, true, SPI_LITTLE_ENDIAN, false);

    //screen initialization
    xScreenDev.spi_dev = (lv_spi_handle_t)&_xSpiDev;
    xScreenDev.protocol = SSD1306_PROTO_SPI3;
    xScreenDev.screen = SSD1306_SCREEN; //SSD1306_SCREEN,  SH1106_SCREEN
    xScreenDev.width    = DISPLAY_WIDTH;
    xScreenDev.height   = DISPLAY_HEIGHT;

    //Reset pin screen initialization
    portable_gpio_enable(PIN_IHM_RST, PORT_GPIO_MODE_OUTPUT);
    portable_gpio_write(PIN_IHM_RST, false);
    vTaskDelay(10 / portTICK_PERIOD_MS);
    portable_gpio_write(PIN_IHM_RST, true);


    //init screen
    while (ssd1306_init(&xScreenDev) != 0)
    {
        debug("%s: failed to init SSD1306 lcd\n", __func__);
        vTaskDelay(1000 / portTICK_PERIOD_MS);
    }
    ssd1306_select_buffer(aucBuffer);

    //inverse screen 180°
    ssd1306_set_scan_direction_fwd(&xScreenDev,true);
    ssd1306_set_segment_remapping_enabled(&xScreenDev, false);

    //Gui initialization
    lv_init();
    _vHAL();

    //Start screen task
    xTaskCreate(_xTaskScreen, "Screen", 512, NULL, screenPRIORITY_TASK, &_xTaskHandleScreen);  //Init TPH ext reading

    //Button task
    xTaskCreate(_Button, "button_task", 512, NULL, screenPRIORITY_TASK, &_xTaskHandleButton);

    //Start timer account for screnn refresh
    _xTimerHandleTick = xTimerCreate("counting_timer", DISPLAY_TICK/portTICK_PERIOD_MS, pdTRUE, NULL, _vTimerTickCounting);
    xTimerStart(_xTimerHandleTick, 100/portTICK_PERIOD_MS);


    debug("Screen system setup done");
}
static void _xTaskScreen(void *pvParameters)
{
    ScreenIcon(); //First screen is the logo

    while(1)
    {
        lv_task_handler();  //Update the buffer
        if(ssd1306_need_redraw())
        {
            ssd1306_load_frame_buffer(&xScreenDev, aucBuffer); // Transmit to the screen the buffer
        }
        vTaskDelay(200 / portTICK_PERIOD_MS);
    }
}
```

kisvegabor:
Much clearer now! Thank you!
However not all details are clear to me we should keep in mind the main goal is the easy porting. When I wrote the already existing drivers (which are not so full-featured then yours) I tried to provide a general super simple interface:
- `xy_disp_init()`;
- `xy_disp_flush()`;
- `xy_disp_map()`;
- `xy_disp_fill()`;

User just need to set these functions for an `lv_disp_drv_t` variable and do the required settings in lv_drv_conf.h
What do you think is it possible to change the API of SSD1306 to look like this? Now it seems not straightforward to me what to do to make the driver work. I see that your implementation mimics the API ESP8266 but I think we should follow a more general way.

Zaltora:
You are maybe right because i don't work on many component yet but maybe you misunderstood/forget something.
I got already theses 4 basic functions. I added the buffer system because lvgl don't support 1 bit/pixel yet and i want save ram. But if i describe a static buffer in ssd1306, i can't use multiple screen like the original library.
What did you prefer ? User select this own buffer or the library get an unique buffer ?(waiting for multi-display support)

After lvgl update we can remove theses functions from the library: 
```
ssd1306_load_frame_buffer();
ssd1306_need_redraw();
ssd1306_select_buffer();
```

The `lv_drv_conf.h` setting describe here is set once for global user porject ;)

kisvegabor:
I see. 
As the multi displays are not supported yet the buffer selection seems to me a little bit confusing.
In my opinion it is enough to prepare the drivers for one display and extend them if required.

And here are two interesting thing:
1. In `dev-5.1` you can set a callback to make rounding on the areas which will be refreshed: `lv_refr_set_round_cb()`. It help to adjust LittlevGL to the needs of SSD1306.
2. In v5.2 it is planned to support "bit level" VDB.

As I see these two feature make the usage of extra buffer unnecessary.

Zaltora:
I wait the release to try this way ? or i use the dev branch ?  I use this driver in my project.

Question: Would be better to add a folder for each different driver ?
It is to improve compilation time and flexibility to get different files for drivers than need it.

kisvegabor:
You will hate me... bu I have an idea :D

A more general API came to my mind which worth a discussion. 
What if we create function like this:
```c
/*Display interface*/
lv_drv_disp_dc_mode(dev, lvl);    /*Data/Command mode: Set D/C pin, add extra bit to SPI or anything you want. You will see the required value.*/
lv_drv_disp_color_mode(dev);      /*You can know that you should use DMA now*/
lv_drv_disp_wr(dev, data, length, word_size);
lv_drv_disp_reset(dev, lvl);
lv_drv_disp_select(dev, lvl);       /*For CS*/

/*Indev interface*/
lv_drv_indev_wr(dev, reg, value);  /*reg: with I2C the reg t owrite, with SPI don't care*/
lv_drv_indev_rd(dev, reg, value);  /*reg: with I2C the reg t owrite, with SPI don't care*/
lv_drv_indev_select(dev, lvl);
lv_drv_indev_rd_irq(dev);         /*Is the touchpad pressed?*/
```

This way we don't need to deal with SPI I2C, Parallel port and GPIO. User can implement is without any constraint. It would give more freedom and clearer API.

What is your opinion?

I'm really sorry that it came to my mind now when we already established a lot of things and you have worked lot. 

To answer your question: I think we should use dev branch until rework the driver to new API. The `master` branch of this repo can be updated independently from `lvgl`.

I thought it not worth to add folders for each driver because they always consist of two files. Or do you see any exceptions?

Zaltora:
I don't understand why this new API can give more freedom. Whatever we use it, driver still need SPI/I2C/GPIO/DMA ...  anyway so we will still use a standard API for theses link.
How mix different protocol usage ?  e.g: SSD1306 can use 3 protocol different and you can do some trick to optimize the way to use them when you develop the driver.

- The user will need to port the protocol from the screen, not only just do a basic porting related to HW interfaces.
- This new api mean to use some extras code (and time) to select the good protocol depending of what was set in "dev".
- Future driver will maybe used a mix of multiple bus/pin.  It more flexible to lets the driver developer to found a way to optimize it ? 
- All driver don't use in the same way the different protocol. How it will be possible to customize them for each driver ? (e.g: SPI for a component will need a specific addr or extra bits depend of he datasheet)
- The API will need to integrate a lot of option and function to cover all possibility. e.g: display can use I2C too and will need "reg" parameter.

IMHO, i think we need just do a abstract layer for each physical link and advanced thing (e.g: DMA). Driver developer need found the best way to mix the different HW API related. What did you think about it ?

I will use dev branch, update and test the changes.

kisvegabor:
You are right. I'm convinced. Thank you for bringing these details to light.

Just for completeness I answer the bullet points:

> The user will need to port the protocol from the screen, not only just do a basic porting related to HW interfaces

For example `lv_drv_disp_wr` is still HW level. Just user can decide which periphery to use to send bytes. I thought the mean freedom ans clearance here because user can decide how or with which periphery  send a byte. 

> This new api mean to use some extras code (and time) to select the good protocol depending of what was set in "dev"

You are right, I missed it with multiple displays. 

> Future driver will maybe used a mix of multiple bus/pin. It more flexible to lets the driver developer to found a way to optimize it?

Can you tell an example?

> All driver don't use in the same way the different protocol. How it will be possible to customize them for each driver ? (e.g: SPI for a component will need a specific addr or extra bits depend of he datasheet)

E.g. `lv_drv_disp_dc_mode` gives to possibility to configure the driver as it is needed.

> The API will need to integrate a lot of option and function to cover all possibility. e.g: display can use I2C too and will need "reg" parameter.

Also true.

Zaltora:
> For example lv_drv_disp_wr is still HW level. Just user can decide which periphery to use to send bytes. I thought the mean freedom ans clearance here because user can decide how or with which periphery send a byte.

The user will need to know how screen work to send the byte : gpio to use, dummy bytes/bits to send, special register to configure before because the driver developer will don't know what the user will use as peripherals. E.g: SSD1306 with i2C need  a extra byte set every x data bytes, SPI don't need.

> E.g. lv_drv_disp_dc_mode gives to possibility to configure the driver as it is needed.

this function it is intended to swap command and data mode.  If we use it for multiple usage to respect protocol, it could become confusing for porting depending on the screens used ?

> Can you tell an example?

[FT81x](http://www.ftdichip.com/Products/ICs/FT81X.html) is a component  input/audio/screen with dual/quad spi and one extras gpio to use.
[RA8876](http://buydisplay.com/download/ic/RA8876.pdf)  Page 10 for diagramm. This component can interface with I2C and SPI and GPIO (button) in same time for different purpose (if i am correct).


PS: theses drivers are very recent and got amazing features for the futures embedded GUI. Take a look on them and tell me what you think ? The RA8876 is a chinese component and look like very awesome.

kisvegabor:
Okay, I already see lv_drv_disp/indev is not a better direction. Thanks!

I know both of the components:
- FT81x: I thing it's a very good product. I wonder why didn't become more popular. I just don't like its interface because it seems to me messy. (Maybe just haven't spent enough time to understand it) And I think its a little bit inconvenient and limited if you can not interact with the GUI directly in the code (e.g. use callback functions). 
- RA8876:  I know this display controller because it has built in GPU. It's also could be a great device (although I never tried it) but it's GPU features cannot be used well in the buffered mode of LittlevGL. In unbuffered mode the `disp_fill` function can be replaced by the RA8876's fill operation but I would rather use the buffered mode with a common display controller via parallel port and DMA.

Zaltora:
Update of SSD1306. I need do some test Wednesday.
- As you said, i will use `lv_refr_set_round_cb()`.
To be sure:
 This function is call once after flush/map/fill process ?
This function is not called if the screen is already refreshed ?

- Added Rst pin in the library, after all, user don't need to manage it. (but need initialize like all peripherals ! )

kisvegabor:
**lv_refr_set_round_cb**
It is called when an area is invalidated in the library. For example when you change the text of a label then you can modify the old and the new area of the text. Or when you change the position of  something the old and the new area is invalidated and can be modified via this callback. So it could be called several times before disp_flush. 
If I understood the things well you can use this function to round every y coordinate to 8. E.g. the area (3,3)(12,22) -> (0,3)(12,23) 

**Reset pin**
Very good idea!

Zaltora:
The main problem is how transfer data to this screen as simply and quickly as possible.
In global situation, you transfer to an area x1,y1 and x2,y2.
For small screen (and monochrome), this method is more difficult that just update the full buffer once.
e.g: a simple text in the middle of the screen is >50% total area.
To do this with lvgl, the best way is to use a VDB in same size that the screen.

I will got a lot of difficulties to change this in SSD1306 librairy:
I got different transfert type and screen protocol. I need rounding and cut the buffer when needed. A lot of work. And i am not sure of well interfaced the static buffer for this version.

the simpler way to do it work, it is to update screen once by loop if needed. It is possible with the current API do to a system like that ?

kisvegabor:
To know how can we use SSD1306 efficiently I need to know how is it really work :)
As I know you can address a column but only every 8th row (or vice-versa). And when you send a byte its bits will be mapped into a column right?

kisvegabor:
... according to the page 25 of SSD1306's datasheet:
image

Zaltora:
yes. 
If i am correct, a good rounding for lvgl update will be like this (to confirm):
(3,3)(22,22) => (3,**0**)(22,**23**)
(35,12)(102,54) =>(35,**8**)(102,**55**)
I suppose (x1,y1) is top/left and (x2,y2) is bot/right. It is always true ?
Here i code i guess can work to update only an area
Before:
For SSD1306:
```c
ssd1306_set_column_addr(dev, 0, dev->width - 1); // Update 0 to max column
ssd1306_set_page_addr(dev, 0, dev->height / 8 - 1); // Update Page 0 to max page
lv_spi_transaction(dev->spi_dev, NULL, _buffer, len, 1); // Send full buffer.
```
After:
```c
//convert y to page
if(y1 != 0) y1/=8;
if(y2 != 0) y2/=8;

ssd1306_set_column_addr(dev, x1, x2); // Update x1 to x2
ssd1306_set_page_addr(dev, y1, y2); // Update y1 to y2

for(u = y1;  u <y2 ; u++) //page
{
    for (i = x1; i < x2; i++) //column
    {
            lv_spi_transaction(dev->spi_dev, NULL, &_buffer[u*dev->width+i], x2-x1, 1);             
    }
}
```
For SH1106:
Before:
```c
for (i = 0; i < (dev->height / 8); i++)
{
     sh1106_go_coordinate(dev, 0, i);
     lv_spi_transaction(dev->spi_dev, NULL, &_buffer[dev->width * i], dev->width, 1);
}
```
After:
```c
//convert y to page
if(y1 != 0) y1/=8;
if(y2 != 0) y2/=8;

for(u = y1;  u <y2 ; u++) //page
{
   sh1106_go_coordinate(dev, 0, u);
    for (i = x1; i < x2; i++) //column
    {
            lv_spi_transaction(dev->spi_dev, NULL, &_buffer[u*dev->width+i], x2-x1, 1);             
    }
}
```
It need to be tested.

kisvegabor:
Yes (x1,y1) is top left, (x2,y2) is bottom right

The codes seems good to me as well. But don't forget that the buffer in `disp_flush` is byte based. So some conversation is required from the "byte map" to a "bit map".
The good thing is it can be done on the spot without any extra buffer.

Zaltora:
I try to update SSD1306 I got some behavior that i need to improve:
bug: i added `#include "lv_core/lv_refr.h"` to lvgl.h or i am unable to use this function. (compilation missing reference)

- i use  `lv_refr_set_round_cb(_rounding_coordinate);`, rounding work ? i am not sure fill/map use new coordinate. Sometimes, this cb is called multiple time for same coordinate ? (See attached files for details)
- function "map" is called for each line update. that mean the screen is update 8 time slower than normal usage. The map function need be called for each page, that mean 8 lines in same time to be perfect.  `lv_refr_set_round_cb` don't round coordinate ?
- function "fill". This one is called for each pixel. This function will update 8 time the same page + send for each pixel coordinate to draw. for the next update, i am not sure if it is possible but it is need update 8 pixel each time to be more efficient.
- function "flush" look work perfectly. For a buffer 1/8 of size screen, it is perfectly rounding and send data only 8 times. The bad thing it is to 8x more memory. If i use a buffer 1/16 of size screen, it send twice the same page. ( the user need config well this if he won't lost performance)

I understand why you want remove map/fill function now. Anyway  map use a buffer to work ( the line) and fill will consume a lot of transmission usage.
I need to use my buffer anyway because map and fill need it. each time to update a pixel already draw, the library need remember it. or each time you will draw a pixel, only 1/8 will be good.
I don't want use VDB to save RAM and i still need extra buffer if the VDB is less than 1/8. When you draw multiple time same area, only last draw will be good. e.g: 1/16 VDB will send 2 time by page, that mean, only 4/8 pixel will be good.

Right now, i can't do better with 5.1 than "hack" the current lvgl system by adding:
```c
if(ssd1306_need_redraw())
{
  ssd1306_load_frame_buffer();
}
```
but i can prepare it by adding the possibility to update only an area. And remove the multi-buffer system that can't be used anyway

PS: I attach log,  image loaded for first screen, then labels write. both situation map/fill and flush. Each line represent a call to a function. Rounding got strange behaviour. called multiple time for same thing. i lets you do analyze on it. 
[log_flush.txt](https://github.com/littlevgl/lv_drivers/files/1822492/log_flush.txt)
[log_map_fill.txt](https://github.com/littlevgl/lv_drivers/files/1822494/log_map_fill.txt)

PS ILI9341: Can you share to me your work on ili9341 with esp32. i will integrate them to my current driver (if possible). if you agree.

Zaltora:
I added last changes on SSD1306 driver. it is a step more to be ready to VDB bit/pixels.

kisvegabor:
Hi,

Yes *rounding* is not working with fill/map properly. And I just realized unfortunately I missed something with flush too... When choosing an area to redraw, it could fail to choose with rounded coordinates. With a screen sized VDB it should work but I will think about how to fix it.

You said you didn't want to use VDB to save memory. On a 128x64 (monochrome) display with 1/8 screen sized VDB it's 1 kB. Is it really worth to save this small amount of RAM and lose all advantages? I mean it's so few RAM. 

I attached the ILI9431 drivers.
[ili9431.zip](https://github.com/littlevgl/lv_drivers/files/1824489/ili9431.zip)

Zaltora:
Hi, i fix a bug on ssd1306  and do a big update for ili9341.
the driver is enough mature to work. only flush is done at 100% using your driver as reference. SPI3 and SPI4 supported but you need do the port for esp32. Can you test it ?
Tell me some feedback if you got time to do it.  ( New API, Organization function)

I use ssd1306 with esp826 and i can tell you, i beg for more memory since weeks XD.
esp-open-rtos original heap memory was 32K. After a lot of different investigation and update, the ram was up to 50K, last week a new update permit to use IRAM as DRAM(20K)(not merged yet) but with limitations. My app use  a lot of memory and i need it in runtime. 1K is so huge for me xD.
Of course i can't tell the same for a esp32 with 512K DRAM ^^.
All advantages from lvgl VDB is not a big lost. It is a small monochrome display, i don't think people need do extraordinary things with it. Some display informations, pictures...
I will continue with memory investigation on ssd1306 to see what can i do to reduce it.

kisvegabor:
Hi,

Thank you very much for the updates. I will test them  as I can. I hope it will be on Thursday.

The RAM thing: I see. :( The problem is with fill/map there is no way to refresh bigger area at once...
However if you allocate a buffer for the screen storing 1 px in 1 bit then it's size is the same as an 1/8 screen sized VDB.

Zaltora:
I thinking about it and you still need use a bit/pixel buffer with VDB. So it always more ram in use :(

kisvegabor:
Hi, 

I checked the updates. It looks very well!

My questions:
1. This comment is still valid in lv_drv_conf_templ.h?
```c
/*
 * All used peripherals must be initialized in user application, library only
 * manipulate them.
 * You can use a device descriptor from your SDK or do your own in this file too.
 * example:
 * typedef struct {...} lv_spi_dev_t;
 */
```
It seems to me `lv_..._dev_t` are replaced by `lv_..._handle_t`

2. In ILI9331 and SSD1306 why `SSD1306_SPI3_SUPPORT` and `SSD1306_SPI4_SUPPORT` need to be defined. I see I2C or SPI need to be set but why SPI3 or SPI4?   
 What about keeping just a `void * `pointer the the `lv_..._handler_t` and the `protocol` field decide how to use it?

Zaltora:
1) i will rework documentation to be more clear. It is was a name suggest for user. handle for the pointer, dev for the object
2) SPI3 or SPI4 are different protocol too and separate. Enable or disable to disable code part on them.
`void*` is the default handle if the user don't need to set a particular one.

e.g for esp8266:
```c
typedef const void* lv_par_handle_t ;
typedef struct {
    uint8_t bus;
    uint8_t cs;
    uint8_t dc;
} lv_spi_dev_t, *lv_spi_handle_t;
typedef const i2c_dev_t* lv_i2c_handle_t ;
typedef uint8_t lv_gpio_handle_t ;
```

Zaltora:
```C
static inline int lv_spi_transaction(lv_spi_handle_t spi_dev, void* data_in, const void* data_out, uint16_t len, uint8_t word_size)
{
	uint32_t data = 0;
	if(!data_out)
	{
		data_out = (void*)&data ;
	}
	if(len == 1 && !data_in)
	{
	    spi_transfer_8(spi_dev->bus, *(uint8_t*)data_out);
	}
	else
	{
	    spi_transfer(spi_dev->bus, data_out, data_in, len, word_size);
	}
	return 0;
}
```

kisvegabor:
Hi,

1. Okay, thank you!
2. SPI3 and SPI4 and their different behavior is an ESP specific thing. Typically an MCU's all SPI periphery are the same.  Why can't the protocol filed decide how to use the handler? If you are on ESP you need to aware of not all SPI knows all protocols.

Zaltora:
I forgot to specify SPI3 and SPI4 are protocol from the screen, not the esp. ^^
you need use SPI in 2 different way.
ILI9341 got  SPI3 and SP4 protocols too with a version I (sda) and a version II (sdi and sdo)
Maybe i need to rename them to Serial3 and Serial4.

kisvegabor:
Oh I see :)
For me `SPI_3_WIRE` and `SPI_4_WIRE` would be clear.

And what about the `void * ` handler and `protocol`?

Zaltora:
ok :)

void* is the default typedef. If you need only transfert a periphals object without interact on him.
e.g:
```c
typedef const void* lv_spi_handle_t;

static inline int lv_spi_transaction(lv_spi_handle_t spi_dev, void* data_in, const void* data_out, uint16_t len, uint8_t word_size)
{
    spi_transfer(spi_dev, data_out, data_in, len, word_size);
    return 0;
}
```

Zaltora:
with esp32, you can keep void* because you can set callback for dc and cs pin directly with the SPI driver.

kisvegabor:
Oh, I seems we misunderstood it. I ment in the code you have:
```c

typedef enum
{
    SSD1306_PROTO_I2C = 0, //!< I2C
    SSD1306_PROTO_SPI4,    //!< SPI 8 bits + D/C pin
    SSD1306_PROTO_SPI3     //!< SPI 9 bits
} ssd1306_protocol_t;

typedef struct
{
    ssd1306_protocol_t protocol;
    ssd1306_screen_t screen;
    union
    {
#if (SSD1306_I2C_SUPPORT)
        lv_i2c_handle_t i2c_dev;         	//!< I2C device descriptor, used by SSD1306_PROTO_I2C
#endif
#if (SSD1306_SPI4_SUPPORT) || (SSD1306_SPI3_SUPPORT)
        lv_spi_handle_t spi_dev;
#endif
    };
    lv_gpio_handle_t rst_pin;
    uint8_t width;                //!< Screen width, currently supported 128px, 96px
    uint8_t height;               //!< Screen height, currently supported 16px, 32px, 64px
} ssd1306_t;
```

My suggestion is:
```c
typedef struct
{
    ssd1306_protocol_t protocol;
    ssd1306_screen_t screen;
   //--- It is changed!!!
    void * drv_handler;         	//!< I2C or SPI driver
    
    lv_gpio_handle_t rst_pin;
    uint8_t width;                //!< Screen width, currently supported 128px, 96px
    uint8_t height;               //!< Screen height, currently supported 16px, 32px, 64px
} ssd1306_t;
```

So just a driver and use it according to `ssd1306_protocol_t protocol;`

Zaltora:
I see. Well we can do in this way too but i prefer use handle to reduce mistake by user when programming.
When you use handle, the compiler will warn to you if the object is not the same. 
with void*, the user need to use a cast.
e.g: the object is a I2C handle, but the protocol selected by user is SPI. You will get a crash.

Of course, he can do the mistake too with the current way but when you see and use a cast "handle_spi_t" with a protocol defined as i2C, you see the problem when writing the code.
What did you think ?

Zaltora:
e.g:
first case:
```c
    xScreenDev.drv_handler = (void *)&_xDev;
    xScreenDev.protocol = SSD1306_PROTO_I2C;
    xScreenDev.screen = SSD1306_SCREEN; //SSD1306_SCREEN,  SH1106_SCREEN
    xScreenDev.width    = DISPLAY_WIDTH;
```
second case:
```c
    xScreenDev.spi_dev = (lv_spi_handle_t)&_xDev;
    xScreenDev.protocol = SSD1306_PROTO_I2C;
    xScreenDev.screen = SSD1306_SCREEN; //SSD1306_SCREEN,  SH1106_SCREEN
    xScreenDev.width    = DISPLAY_WIDTH;
    xScreenDev.height   = DISPLAY_HEIGHT;
```
The error is obvious is the second case.

kisvegabor:
You are right. You will get an error on warning. The only thing is now you see an `spi_dev` and an `i2c_dev` which might be confusing.

However with a proper commenting in the `struct` your solution is fine!

PR #4, opened 2018-03-13
Fixed extern C guards in all c-style headers

memen45:


kisvegabor:
Great!
Thank you very much!

PR #7, opened 2018-04-01
fix ssd1306 + ILI9341 transmissions upgrade

Zaltora:
Hey, i update ili9341 driver and do a small fix on ssd1306.
Ili9341 update mainly update transmission system. flush/map/fill are complete and two major changes are here:
- Endian : As you know, you need inverse byte order when you send pixel to the screen, but some SPI driver can do it without need additional treatment. So i added a option for user to swap byte order or not if the SPI support it. (esp8266 support it, esp32 too if i am right)
- While/for : I got some infinite loop bug when setting was wrong ? i guess  that size calculation in certain case is not round... whatever. Using for loop is more robust (still guess) .

I got a screen ILI9341 with a electronic card from 4Dsystem. Test done. Everything work well !! byte order, algorithm, different VDB size. Feedback welcomed. Got a visual bug with fill/map but i think it is lvgl option that is wrong.
Next stage will be for doing a driver for AR1021 touch controller to finish test on the electronic card. If you know this controller or you got a driver for it, tell me. thank.

### DIFF
--- a/display/ILI9341.c
+++ b/display/ILI9341.c
@@ -120,7 +120,6 @@
 #define ILI9341_ENABLE_3G               (0xF2)
 #define ILI9341_PUMP_RATIO_CTR          (0xF7)
 
-
 /**********************
  *      TYPEDEFS
  **********************/
@@ -130,7 +129,7 @@
  **********************/
 static int _sendCommand(const ili9341_t *dev, uint8_t cmd);
 static int _sendCommandData(const ili9341_t *dev, uint8_t cmd, uint8_t* data_out, uint32_t len);
-static int _sendData(const ili9341_t *dev, uint8_t* data_out, uint32_t len);
+static int _sendData(const ili9341_t *dev, uint8_t* data_out, uint32_t len, uint8_t wordsize);
 static int _receiveData(const ili9341_t *dev, uint8_t cmd, uint8_t* data_in, uint32_t len);
 
 /**********************
@@ -143,7 +142,7 @@ static ili9341_t* _device; //FIXME: To remove when multi-display will be support
  **********************/
 #if ILI9341_DEBUG
 #include <stdio.h>
-#define debug(fmt, ...) printf("%s: " fmt "\n", "ILI9341", ## __VA_ARGS__)
+#define debug(fmt, ...) printf("%s: " fmt "\n", __FUNCTION__, ## __VA_ARGS__)
 #else
 #define debug(fmt, ...)
 #endif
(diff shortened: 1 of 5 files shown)


Zaltora:
- If ili9341 got extended command enable, you can change endian setting input for the screen. That mean don't need to swap color byte. 
- Possible to set 262K color mode using color24, It is work but i am not sure if i am correct for send color. bit shifting is needed  but when i do it, i got some problem, and when i don't use it i don't get any problem. I don't see the difference visually too.
- fix a big mistake for error management, if error was detected in spi mode, the line was not correctly released. Now it will be released after each transaction. Other advantage, when you use multi-task system, the system can swap task during big data transfer to the screen. Disadvantage, you will lost a little performance if you use SW CS/DC pin.
- rename some things and fix

Zaltora:
Update the way to send pixel to the screen:
 - fill/map/flush :  cut buffer into smaller chunk if needed, defined by user with "ILI9341_MAX_SAMPLE"
 - Format pixel to send like swap byte or special bit format depending of protocol used and color.
 - Send data 
fix typos and initial porting for parallel protocolusage (WIP).

kisvegabor:
Wao!  These are very great updates! 
- I really like the ILI9341 byte swap feature. I didn't know it. 
- I didn't used AR1021 yet. :(

i will check the updates in more detail soon.

Zaltora:
it is exist 3 way to solve byte order for  65K color mode: CPU swap, SPI setting and ILI9341 setting ( if EXTC pin is high). I  Can't do test on the third solution because i don't get extended mode command.
I discovered that a lot of command don't need to be send at initialization start when the pin EXTC is not high.

I think about a problem. When i get different spi frequency for different device, what to do ?. Check if frequency is correct for each transaction and don't need extras API (depend of SDK anyway) or add a new function prototype to change frequency on demand when write the driver. I prefer first solution, i will try it. what you think ?

I think about add a core file protocol data format. Many protocols and data transformation are common between screen driver. When i will finish the ILI9341 parallel port, it will be easier to do the SSD1963. Protocols are the same.

I think about add PWM portable API for backlight, and create an API standard for each screen like:
-(driver)_set_backlight();
-(driver)_set_rotation();
-(driver)_inverse_color();
...

I will do the AR1021 driver from start then :)

kisvegabor:
**SPI  speed**
Good remark. My goal is typically to make the APIs as simple as possible. So I also vote for user check the speed if needed.

> I think about add a core file protocol data format.

Can you explain it in more detail? 

**Extra API function**
Very good idea. It's very common to set the backlight and rotation. IMHO inverse color is relevant only is special cases with monochrome display as we discussed it earlier. 
Maybe the functions can be registered in `lv_disp_drv_t`. It would be useful with multi-displays because this way a display would be described with a structure. For example when you use a slider to set the brightness you can get the object's display and call its brightness function.

PR #9, opened 2018-04-09
Ar10xx driver + others fixes

Zaltora:
New driver incoming:
- AR1010, AR1011, AR1020 and AR1021 indev driver (WIP)
- fixes typos for ili9341
- fix I2C transmission for SSD1306

AR10XX not tested yet. Current implementations:
-EEPROM user acces.
-Basics Command (id, save,load, read, write,..)
-All register configurations
-I2C and SPI protocol 
-Feedback from command (IRQ/delay)

Todo:
-Basics reading coordinate function + IRQ function
-UART implementation
-Lvgl interupt read update system
-buffered configuration, save/load configs into a struct.

### DIFF
--- a/display/ILI9341.c
+++ b/display/ILI9341.c
@@ -458,6 +458,23 @@ int ili9341_init(ili9341_t *dev)
 }
 
 
+int ili9341_set_rotation(ili9341_t *dev, ili9341_rotation_t degree)
+{
+    switch(degree)
+    {
+        case ILI9341_DEGREE_0:
+            break;
+        case ILI9341_DEGREE_90:
+            break;
+        case ILI9341_DEGREE_180:
+            break;
+        case ILI9341_DEGREE_270:
+            break;
+        default:
+            break;
+    }
+}
+
 int ili9341_unknow(const ili9341_t *dev)
 {
     uint8_t data[3] = { 0x03, 0x80, 0x02  };
@@ -919,9 +936,9 @@ static int _sendDataPixels(const ili9341_t *dev, const lv_color_t *pixel, uint32
 {
 #if ILI9341_DEBUG
     printf("%s: ",__FUNCTION__);
-    for (uint8_t i = 0 ; i < len ; i++)
+    for (uint16_t i = 0 ; i < len ; i++)
     {
-        printf("%02X ",data_out[i]);
+        //printf("%02X ",pixel[i]);
     }
     printf("\n");
 #endif
--- a/display/ILI9341.c
+++ b/display/ILI9341.c
@@ -15,18 +15,6 @@
 /*********************
  *      DEFINES
  *********************/
-#ifndef ILI9341_SPI3WIRE_SUPPORT
-#define ILI9341_SPI3WIRE_SUPPORT (1)
-#endif
-
-#ifndef ILI9341_SPI4WIRE_SUPPORT
-#define ILI9341_SPI4WIRE_SUPPORT (1)
-#endif
-
-#ifndef ILI9341_PAR_SUPPORT
-#define ILI9341_PAR_SUPPORT (1)
-#endif
-
 #ifndef ILI9341_COLOR
 #define ILI9341_COLOR (65)
 #endif
--- a/display/ILI9341.c
+++ b/display/ILI9341.c
@@ -140,16 +140,6 @@ typedef struct {
 /**********************
  *  STATIC PROTOTYPES
  **********************/
-#if (ILI9341_SPI3WIRE_SUPPORT)
-static int inline _spi3wire_send(const ili9341_t *dev, bool dc, uint8_t* data, uint8_t len, uint8_t wordsize);
-#endif
-#if (ILI9341_SPI4WIRE_SUPPORT)
-static int inline _spi4wire_send(const ili9341_t *dev, bool dc, uint8_t* data, uint8_t len, uint8_t wordsize);
-#endif
-#if (ILI9341_PAR_SUPPORT)
-static int inline _par_send(const ili9341_t *dev, bool dc, uint8_t* data, uint8_t len, uint8_t wordsize);
-#endif
-
 static int _sendCommand(const ili9341_t *dev, uint8_t cmd);
 static int _sendCommandData(const ili9341_t *dev, uint8_t cmd, uint8_t* data_out, uint32_t len);
 static int _sendDataPixels(const ili9341_t *dev, const lv_color_t *pixel, uint32_t len, uint8_t wordsize);
@@ -284,8 +274,9 @@ void ili9341_map(int32_t x1, int32_t y1, int32_t x2, int32_t y2, const lv_color_
 
 /* Perform default init routine according */
 //TODO: don't forget to make it constant again
-int ili9341_init(ili9341_t *dev)
+int ili9341_init(ili9341_t *dev, lv_rotation_t rotation)
 {
+
     if(_device)
     {
        return -1; //A screen was already initialized.
@@ -302,7 +293,6 @@ int ili9341_init(ili9341_t *dev)
         debug("Unsupported screen height");
         return -ENOTSUP;
     }
-
     if(dev->rst_pin != LV_DRIVER_NOPIN)
     {
         lv_gpio_write(dev->rst_pin, 0);
@@ -321,10 +311,15 @@ int ili9341_init(ili9341_t *dev)
     //if((err = ili9341_vert_scroll_start(dev, ili9341_vert_scroll_start_config)))
     //    return err;
 
-    ili9341_mem_ctrl_t ili9341_mem_ctrl_config = { 0 } ;
-    ili9341_mem_ctrl_config.mx = 1 ;
-    ili9341_mem_ctrl_config.bgr = 1 ;
-    err_control(ili9341_mem_ctrl(dev, ili9341_mem_ctrl_config));
+    //ili9341_mem_ctrl_t ili9341_mem_ctrl_config = { 0 } ;
+    //ili9341_mem_ctrl_config.mx = 1 ;
+    //ili9341_mem_ctrl_config.bgr = 1 ;
+    //err_control(ili9341_mem_ctrl(dev, ili9341_mem_ctrl_config));
+
+    dev->r = 0;
+    ili9341_set_rotation(dev, rotation);
+    dev->r = rotation;
+
 
     ili9341_px_fmt_t ili9341_pixel_fmt_config = { 0 };
 #if (ILI9341_COLOR == 65)
@@ -458,21 +453,39 @@ int ili9341_init(ili9341_t *dev)
 }
 
 
-int ili9341_set_rotation(ili9341_t *dev, ili9341_rotation_t degree)
+int ili9341_set_rotation(ili9341_t *dev, lv_rotation_t degree)
 {
-    switch(degree)
+    int8_t rotate = (dev->r + degree)%4;
+
+    ili9341_mem_ctrl_t config = { 0 };
+    switch(rotate)
     {
-        case ILI9341_DEGREE_0:
+        case LV_ROT_DEGREE_0:
+            config.mx = 1;
+            config.bgr = 1;
+            ili9341_mem_ctrl(dev, config);
             break;
-        case ILI9341_DEGREE_90:
+        case LV_ROT_DEGREE_90:
+            config.mx = 1;
+            config.mv = 1;
+            config.my = 1;
+            config.bgr = 1;
+            ili9341_mem_ctrl(dev, config);
             break;
-        case ILI9341_DEGREE_180:
+        case LV_ROT_DEGREE_180:
+            config.my = 1;
+            config.bgr = 1;
+            ili9341_mem_ctrl(dev, config);
             break;
-        case ILI9341_DEGREE_270:
+        case LV_ROT_DEGREE_270:
+            config.mv = 1;
+            config.bgr = 1;
+            ili9341_mem_ctrl(dev, config);
             break;
         default:
             break;
     }
+    return 0;
 }
 
 int ili9341_unknow(const ili9341_t *dev)
@@ -849,62 +862,6 @@ int ili9341_set_page_addr(const ili9341_t *dev, int32_t start, int32_t stop)
 /**********************
  *   STATIC FUNCTIONS
  **********************/
-#if (ILI9341_PAR_SUPPORT)
-static int inline _par_send(const ili9341_t *dev, bool dc, uint8_t* data, uint8_t len, uint8_t wordsize)
-{
-    lv_par_wr_cs(dev->spi_dev, false);
-    lv_par_wr_dc(dev->spi_dev, dc); /* command mode */
-#if (ILI9341_ERR_CHECK)
-    int err = lv_par_write(dev->spi_dev, data, len, wordsize);
-#else
-    lv_par_write(dev->spi_dev, data, len, wordsize);
-#endif
-    lv_par_wr_cs(dev->spi_dev, true);
-#if (ILI9341_ERR_CHECK)
-    return err;
-#else
-    return 0;
-#endif
-}
-#endif
-#if (ILI9341_SPI3WIRE_SUPPORT)
-static int inline _spi3wire_send(const ili9341_t *dev, bool dc, uint8_t* data, uint8_t len, uint8_t wordsize)
-{
-    lv_spi_wr_cs(dev->spi_dev, false);
-    lv_spi_set_preemble(dev->spi_dev, LV_SPI_COMMAND, dc, 1);
-#if (ILI9341_ERR_CHECK)
-    int err = lv_spi_transaction(dev->spi_dev, NULL, data, len, wordsize);
-#else
-    lv_spi_transaction(dev->spi_dev, NULL, data, len, wordsize);
-#endif
-    lv_spi_clr_preemble(dev->spi_dev, LV_SPI_COMMAND);
-    lv_spi_wr_cs(dev->spi_dev, true);
-#if (ILI9341_ERR_CHECK)
-    return err;
-#else
-    return 0;
-#endif
-}
-#endif
-#if (ILI9341_SPI4WIRE_SUPPORT)
-static int inline _spi4wire_send(const ili9341_t *dev, bool dc, uint8_t* data, uint8_t len, uint8_t wordsize)
-{
-    lv_spi_wr_cs(dev->spi_dev, false);
-    lv_spi_wr_dc(dev->spi_dev, dc); /* command mode */
-#if (ILI9341_ERR_CHECK)
-    int err = lv_spi_transaction(dev->spi_dev, NULL, data, len, wordsize);
-#else
-    lv_spi_transaction(dev->spi_dev, NULL, data, len, wordsize);
-#endif
-    lv_spi_wr_cs(dev->spi_dev, true);
-#if (ILI9341_ERR_CHECK)
-    return err;
-#else
-    return 0;
-#endif
-}
-#endif
-
 static int _sendCommand(const ili9341_t *dev, uint8_t cmd)
 {
     debug("cmd: %02X",cmd);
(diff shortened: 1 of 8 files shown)


kisvegabor:
Thank you. 
Can you also test the AR10XX drivers?

Zaltora:
yes i got a AR1021 touch controller, i will test it end of this week.

kisvegabor:
Then you for the update!

Is it still work in progress or shall I merge it?

Zaltora:
In progress. I will do test soon. I will prevent you. :)

Zaltora:
I can't test I2C and i think it will be difficult to port this point.
datasheet:  A 50 us delay is needed before the Stop bit.
Clock stretching not supporting in writing mode :(.

I will continue test next week, i got a cable broken. It is close to work for SPI.

kisvegabor:
Hi,

Oh, what a strange  expectation. I never heard anything like this before.
I think it's okay it will work with SPI for now.

Zaltora:
Okay, i added an error if someone want use AR1020 in I2C mode. Doing a lot of test. Touch controller work very well ! i fixed a lot of mistake and do some rework on "how to receive data" with timeout and small delay ( for system that need sleep or get watchdog). Tested the possibility to write on the EEPROM, load config, save, ... All feel to work. :)

The next big dev stage is how to calibrate the touch controller and sync it with the screen.
I will introduce the rotation system to prepare the possibility to change the orientation dynamically like phone. It will be done with the rotation API and 4 state (0,90,180,270 degree).
device will need to know this:
x1,y1,x2,y2, height, width, rotation.

It is in WIP,  but the current dynamic calibration work. I need finish and enchancce it. The API to do:

```c
/**
 * Wait a event on the screen to save current coordinate as reference
 * @param dev Pointer to the touch controller descriptor
 * @param stage the touch location to do ( screen corner).
 * @param number Number of screen touch to do
 * @param max_delay Timeout in second
 * @return Non-zero if error occurred
 */
int ar10xx_map_screen_coordinate(ar10xx_t *dev, ar10xx_calib_t stage, uint8_t number, uint16_t max_delay);

/**
 * Put touch controller into calibration mode and touch each corner of the screen.
 * It will calibrate the screen output data (12bits)
 * @param dev Pointer to the touch controller descriptor
 * @param max_delay Timeout in second
 * @return Non-zero if error occurred
 */
int ar10xx_screen_calib(ar10xx_t *dev, uint16_t max_delay);

/**
 * Change the coordinate map to correspond to the current rotation
 * @param degree current rotation
 * @return Non-zero if error occurred
 */
int ar10xx_screen_rotation(ar10xx_t *dev,  ar10xx_rot_t degree);
```
When you init the screen, you can set the default calibration with define. Other thing to think is how to do a good calibration. Each corner is difficult. I think about add a parameter like pixel offset.
To work properly, we need a simple screen function to put a target for the user.
and the target will change each time a touch is detected. How you think we can do this ?

I think about a special lvgl object linked to screen standard API and show target for user and with few setting like corner (topleft,topright,...) offset, timeout and touch number.
the touch controller API need be called before showing a target to the screen and will return error if nobody touch this screen before delay timeout?
Or maybe you know already a way to do it :)

kisvegabor:
Hi,

Looks great!

Just two question about the API:
-  wouldn't be better to change `ar10xx_rot_t` and `ar10xx_calib_t` to "global types"? Like `lv_disp_rotate_t` and `lv_indev_claib_state_t`. Just because They are the same for every screen and touchpad type.
- To separate the driver from the GUI I can imagine a function like `ar10xx_set_calib_data()` instead of `ar10xx_map_screen_coordinate` and `ar10xx_screen_calib`. This way we can have a touch pad calibration GUI where we can pas the coordinates to the touchpad driver. 

There is a Touchpad calibration application: https://github.com/littlevgl/lv_examples/tree/master/lv_apps/tpcal

It registers the 4 coordinates of the 4 corners. The application can be updated/modified the fulfill or needs. 

For calibration I usually use 4 data (x1,x1,y1,y2) of an uncalibrated touchpad:
- x_cal = (x_raw - x1) / (x2-x1) * x_res;
- y_cal = (y_raw - y1) / (y2-y1) * y_res;

What do you think?

PS:
I will be on a holiday for a next Sunday (29th of April) and probably I won't be able to answer to the emails.

Zaltora:
hi,
- i agree to get a general rotation struct (it was already in my mind)
- Yes, i will use your example to pass  the coordinate to the driver. I will found a way.
- I will use this formulas too :)

I will get tme soon to work on it.

kisvegabor:
Awesome! :)

For the formula is better to change the order of the operations for higher precision:
`x_cal = ((x_raw - x1)  * x_res) / (x2-x1);`

Zaltora:
Hi, i am back to work on lvgl :). I will finish rotation and calibration features.
Then i will so some clean up and documentation before go to next driver.

I got a problem with example "tpcal" and i need some help to understand why.
I can got for a corner an invalid number. depend where i touch. But when i look the input data, i didn't see invalid number.
Can be a problem with return false/true ?
Or maybe that the last input information was not released data (p: 0)

e.g:  the result when i touch screen (top corner and right corner)
AR10XX input result:
```
--(1)
ar10xx_input_get: p: 0, x: 5, y: 9
ar10xx_input_get: p: 1, x: 6, y: 9
ar10xx_input_get: p: 1, x: 6, y: 9
--(2)
ar10xx_input_get: p: 0, x: 227, y: 13
ar10xx_input_get: p: 1, x: 227, y: 13
ar10xx_input_get: p: 0, x: 228, y: 13
```
screen corner result:
```
--(1)
x =  -31024
y = 16383
--(2)
x = 228
y = 13
```

the C Read function:
```c
bool ar10xx_input_get(lv_indev_data_t * data)
{
    /*Get coordinate*/
    ar10xx_read_t position;
    if (_read_pos(_device, &position))
    {
        return false;
    }
    /*Store the collected data*/
    data->point.x = (((int32_t)position.x - (int32_t)_device->x1) * (int32_t)_device->w) / ((int32_t)_device->x2 - (int32_t)_device->x1);
    data->point.y = (((int32_t)position.y - (int32_t)_device->y1) * (int32_t)_device->h) / ((int32_t)_device->y2 - (int32_t)_device->y1);
    data->state = position.pen ? LV_INDEV_STATE_PR : LV_INDEV_STATE_REL; //Depend setting

    debug("p: %u, x: %d, y: %d", data->state, data->point.x, data->point.y);
    return false;
}
```

Calibration:
I  think  a way to calibrate the input data range  ( raw_x and raw_y).

My first approach was to call a function `ar10xx_map_screen_coordinate` (will be renamed). This function block until someone touch the screen. When touched, the system will get raw limit x,y.
This limit is adjusted with an offset which is calculated based on the size of the circle displayed on the screen. (offset = the radius of the circle)

If i understand well the tpcal example, it is work like that:
- you register driver reading function input
- you get point from this function (already screen sized computed) with the driver
- then you store this point in a array  `p[4]`.

This is not accurate if we can not retrieve the raw information of the sensor? an idea ?
My idea was to save the raw coordinate each time the screen is touched.
Use `ar10xx_set_calib_data` will get the raw coordinate and save them.

kisvegabor:
Hi,

Welcome back! :)

It's very strange because the `lv_indev` saves the data directly from  `lv_indev_data_t`.
Only the first point is wrong?

About the `tpcal` workflow:
Not exactly. I imagined it like this: 
1. You display driver should have two functions: `xxx_read` and `xxx_set_cal_data()`
2. At first run the TP is not calibrated therefore your `read` function should provide raw data
3.When you touched all four corners you should call the `xxx_set_cal_data` function in `TP_CAL_STATE_WAIT_LEAVE` state of `tpcal`.
4. `xxx_set_cal_data` should mark the display as calibrated, save the raw coordinates, and use them in the upcoming calls of `xxx_read`

Well it is not well documented. I will add these these steps to `tpcal` to inform users about the intended usage.

Did it help?

Zaltora:
Hi, okay then i got 2 function
=> `bool ar10xx_input_get_raw(lv_indev_data_t * data);`
=> `bool ar10xx_input_get_calib(lv_indev_data_t * data);`

When i calib, i register raw input function. after calib, i register the calib function for normal usage.
After getting point, i call this function
`int ar10xx_set_calib_data(ar10xx_t *dev, lv_point_t* pts, int16_t offset);`
It take the array ( order is important. T/L, T/R, B/R, B/L) and a offset border.
The offset border, this information is usefull because the user will target the center of a circle. that mean, he will touch the screen with a offset of some pixel. I use it to determine the raw input at true border because i guess that the touch get linear input function.

I am ready to test it but `tpcal` is buggy. All points is concerned, and after some test i can tell you. If the last point read from input don't get the released state (p). It will fail.
Last input trig, e.g:
`ar10xx_input_get: p: 1, x: 6, y: 9`  => fail and get (x:-31024 and y:16383)
`ar10xx_input_get: p: 0, x: 6, y: 9`  => succes and good coordinate

my call in tpcall function (only 3 line):

```c
        /*
         * TODO Process 'p' points here to calibrate the touch pad
         */
        ar10xx_set_calib_data(&xInputDev, p, CIRCLE_SIZE/2);
        lv_indev_t *indev = lv_indev_get_act();
        indev->driver.read = ar10xx_input_get_calib; //normal usage now
```

Zaltora:
Improvement `tpcal`:
-Larger circle or offset  circle: to be easier to touch for a person, the algorithm will determine true border. 
-Multiple touch : Touch multiple time the same location for an average.

Zaltora:
Question about rotation API.How we will work to manage rotation
-use int16_t and use number in degree or use enumerator ?
-use negative rotation ?

e.g AR10XX rotation function:
```c
void ar10xx_set_rotation(ar10xx_t *dev,  int16_t degree)
{
    degree = degree / 90 ;

    //+/-180 degree rotation
    if ((degree == 2) || (degree == -2))
    {
        SWAP(dev->x1, dev->x2)
        SWAP(dev->y1, dev->y2)
    }
    //-90 ou +270 degree rotation (left to right)
    else if ((degree == -1) || (degree == 3))
    {
        int16_t temp = dev->x1;
        dev->x1 = dev->y2;
        dev->y2 = dev->x2;
        dev->x2 = dev->y1;
        dev->y1 = temp;
        SWAP(dev->w, dev->h);
    }
    //+90 ou -270 degree rotation (right to left)
    else if ((degree == 1) || (degree == -3))
    {
        int16_t temp = dev->x1;
        dev->x1 = dev->y1;
        dev->y1 = dev->x2;
        dev->x2 = dev->y2;
        dev->y2 = temp;
        SWAP(dev->w, dev->h);
    }
}
```
this example use relative rotation, not absolute. It mean that call this function 2 time with 90 degree is like call this function with 180.
But i don't no if we work like this or just use absolute rotation ?

bay-zoltan-nonprofit-ltd:
Hi,

1. The **offset parameter** is a good idea. You are right, user targets the center of the circle which would result an error.

2. Using the `CIRCLE_SIZE` define it's easy to increase circle's size. 
The average calculation also sounds great. 

3. I have no idea why do you see wrong coordinates. I already used this app in a real life project and it worked well. So you say always the last coordinate is wrong is no release happen?

4. For me, in rotation API, using enums sounds the best option. Like: `LV_INDEV_ROTATE_90`.
By the way, it happened to me a lot of time that the x or y coordinates were inverted (e.g. the left most coordinate was the 2048 and right most was 0) or the x-y coordinates was swapped (x was the vertical coordinate).
I think it should can be implemented in LittlevGL itself because these corrections are the same for all touchpad like devices. That reminds me that the rotation can be part of the internal indev handling as well. It's just a parameter of `lv_indev_t` and LittelvGL can swap/invert/etc the coordinates in a few lines of code. Maybe adding some API function like: `lv_indev_set_orientation()` or `lv_indev_set_xy_inversion()`. This was the drivers can be more simple. What do you think?

kisvegabor:
Sorry, the last was also me accidentally with an other account :)

Zaltora:
1. Automatic calibration function work perfectly. Tested with different radius. calculation seem good.

2. I think, we need to add a small cross in the circle center too. ( to help )

3. Whatever the corner you push, if the input device read function don't send `LV_INDEV_STATE_REL` as last information for lvgl, it will fail ! I use "dev-5.2" branch (last update today)
debug the input information in your project. If you got always the good information, it will work nicely. Not my case :( . It is very strange. My reading function :
```c
bool ar10xx_input_get_raw(lv_indev_data_t * data)
{
    /*Get coordinate*/
    ar10xx_read_t position;
    if (_read_pos(_device, &position))
    {
        return false;
    }
    data->point.x = position.x;
    data->point.y = position.y;
    data->state = position.pen ? LV_INDEV_STATE_PR : LV_INDEV_STATE_REL;
    return false;
}
```
Clue: The x and y got strange value when they are not set. That mean when `_read_pos` is true. the function will return without set the pointer "data".
This function fix the problem. Ugly:
```c
bool ar10xx_input_get_raw(lv_indev_data_t * data)
{
    static int16_t x = 0;
    static int16_t y = 0;
    static int16_t p = LV_INDEV_STATE_REL;

    /*Get coordinate*/
    ar10xx_read_t position;
    if (_read_pos(_device, &position))
    {
        data->point.x = x;
        data->point.y = y;
        data->state =  p;
        return false;
    }
    x = position.x;
    y = position.y;
    p = position.pen ? LV_INDEV_STATE_PR : LV_INDEV_STATE_REL;
    data->point.x = x;
    data->point.y = y;
    data->state = p;
    return false;
}
```
4. WIP with your new information.

Zaltora:
In fact i got new problem with the new function. when the input device decide to send twice the released state, it like push 2 time.

Input device signal order when you physically touch the device:
case 1:
```
p:0
p:1
...
p:1
p:0
```
case 2:
```
p:0
p:1
...
p:1
p:1
```

Zaltora:
I fixed all problem when i force last input read to be a released state:
```c
bool ar10xx_input_get_raw(lv_indev_data_t * data)
{
    static int16_t x = 0;
    static int16_t y = 0;
    static int16_t p = LV_INDEV_STATE_REL;
    /*Get coordinate*/
    ar10xx_read_t position;
    if (_read_pos(_device, &position))
    {
        if(p) //Protection if last input read was pressed state
        {
            p = LV_INDEV_STATE_REL;
            data->point.x = x;
            data->point.y = y;
            data->state =  p;
        }
        return false;
    }
    x = position.x;
    y = position.y;
    p = position.pen ? LV_INDEV_STATE_PR : LV_INDEV_STATE_REL;
    data->point.x = x;
    data->point.y = y;
    data->state = p;
    return false;
}
```
Perhaps we need to review the use of the return of the function.
I tested `tpcal` with "return true" whenever the driver reading is a success.
`tpcal` considers each reading as a new input corners. funny behavior ^^

Zaltora:
4. Here is my thought on the rotation system:

When initialize the driver you need specify 3 things. Width, height and rotation.
At startup, it is allow the rotation of the screen and touchpad to be different. if the touchpad was put differently for example.
```c
driver_indev_init (dev, 320, 240,  LV_ROT_DEGREE_90);
driver_display_init( dev, 320, 240, LV_ROT_DEGREE_180);
```
then we can register 2 function in lvgl, one for indev, one for display
these function will turn the screen from origin.
```c
 void lv_indev_set_orientation( lv_rotation_t orientation);
 void lv_display_set_orientation( lv_rotation_t  orientation);
```
when user will use lvgl API, it will  look like this:
```c
void  lv_set_orientation( lv_rotation_t orientation);
```
Of course, we need link a indev with a display. We need remove `LV_HOR_RES` and  `LV_VER_RES` macros (need to be dynamic).
What did you think? (commit soon if you want take a look)

kisvegabor:
Hi,

Thank you for the feedback and great work!

**Read function**
I usually use this structure in my read functions:
```c
bool xpt2046_read(lv_indev_data_t * data)
{ 
    static int16_t last_x = 0;
    static int16_t last_y = 0;
    bool valid = true;
    
    int16_t x;
    int16_t y;
 
   valid = xxx_tp_get(&x, &y);
   if(valid) { 
       last_x = x;
       last_y = y;
    } else {
        x = last_x;
        y = last_y;
    }
    
    data->point.x = x;
    data->point.y = y;
    data->state = valid == false ? LV_INDEV_STATE_REL : LV_INDEV_STATE_PR;

    return false;
}
```
So far it worked me in every cases. As LittelvGL waits for valid coordinates after every call of `read` you should save the last valid coordinates.

**Orientation**
Your idea sound great! The multi display system would require the removal of `LV_HOR_RES` and `LV_VER_RES`. At first look it could cause a massive API change. However my plan is to add multi display support in a backward compatible manner. This way LV_HOR_RES/VER_RES can be only initial values of a one-displayed non rotated configuration, but you might have the opportunity to overwrite this settings. I need to examine it deeper.

Zaltora:
Okay i take the notes about read function. But what is the purpose to return "false" or "true" ? seem like we never use true.
A question, at tpcal start, i want do this:
```c
    lv_indev_t *indev = lv_indev_get_act();
    if(indev)
    {
        indev->driver.read = ar10xx_input_get_raw;
    }
```
but it is return NULL. The input driver was register in lvgl just before. 
Why the actual device can't be the last input registered/enable ? 

You can merge this PR !!
The main logic about calibration and rotation is done !! algorithm work well, need just update `tpcal` with average to reduce the error.
calibration algorithm need maybe a review to be optimized.
AR10XX driver work well, calibration and rotation same. I am limited for the last point because i can't turn display of 90 degree. ( h and w need change different into lvgl).

The next major PR will be about  cleanup / documentation / code structure /  fixes?.
No need to hurry for the multi-display. Still need some drivers to do and some validation.
The bit per pixels update too :). that will allow me some additional clean-up.

kisvegabor:
**read return value**
If returns with `true` then you indicates that you still have some buffered data to read. In this case LittelvGL continue to read the device until get `false`.

**lv_indev_get_act**
`lv_indev_get_act` return the a pointer of the active indev. The active indev is the one which called an action or callback. So for example you can determine in a button action which indev called it, and e.g. can read the pressed point.

**Orientation**
Sorry, I'm afraid I missed something and forgot to discuss this: Earlier I mentioned the indev orientation can be handled internally in LittlevGL (without any modification and extension in the driver) with swapping/inverting the coordinates internally in the library. However it's not simple with the display. So my view is to keep the `lv_display_set_orientation` function but the indev orientation for LittelvGL. The initial orientation adjustment still required to set it according to the current hardware configuration.  What do you think?

Zaltora:
1) I found maybe a bug when using `tpcal` then. When i change raw input read with this:
```c
bool ar10xx_input_get_raw(lv_indev_data_t * data)
{
    /*Get coordinate*/
    ar10xx_read_t position;

    if (_read_pos(_device, &position))
    {
        if(_device->p == LV_INDEV_STATE_PR) //Protection if last input read was pressed state
        {
            _device->p = LV_INDEV_STATE_REL;
            data->point.x = _device->x;
            data->point.y = _device->y;
            data->state =  _device->p;
        }
        return false;
    }

    _device->x = position.x;
    _device->y = position.y;
    _device->p = position.pen ? LV_INDEV_STATE_PR : LV_INDEV_STATE_REL;
    data->point.x = _device->x;
    data->point.y = _device->y;
    data->state = _device->p;

    return true; //Return true this time to see what happen
}
```
lvgl consider in `tpcal` each input read as a new circle touch point when i put my finger and didn't move.
The demo are not affected by this change (visually)

2) If i understand well, the input dev need to be used once before be considered as active ?
I want that `tpcal` example change the driver input function callback before begin to retrieve points. How i can do this ?

3) You are right. It can be a lvgl layer. I got an idea. I created in this PR a new file that got common things.
lvgl don't support rotation yet a guess ? I keep this features to make driver work on current lvgl version, then i remove unnecessary code as lvgl update ?
i will do some changes and think about what i need to do.

kisvegabor:
Hi,

1. The tpcal creates a very big button. When you press the screen, really you are pressing/releasing this big button. `btn_click_action` will be called when you release the screen. So you can move your finger, but nothing will happen until you release the touchpad.

2. Not exactly. Imagine that you have 2 input devices: a mouse and a touch pad. While lvgl processing the data of the mouse the mouse is active. After it while it processes the data of the touch pad the touchpad indev is active. When both are ready none of them is active. 
If you have only one indev you can use: `lv_indev_t * indev = lv_indev_next(NULL)`. But if you have (or might have) more indevs you should create a function yourself, like: `get_my_touchpad_indev()` or use a global variable. 

3. Sounds very well! You are right, now lvgl doesn't support rotation feature but I can add it v5.2.

Zaltora:
>  So you can move your finger, but nothing will happen until you release the touchpad.

that the problem. the behavior that you describe don't happen.
Everything happen without need release information.
video: [VID_20180605_174433.zip](https://github.com/littlevgl/lv_drivers/files/2073064/VID_20180605_174433.zip)

kisvegabor:
It works for as expected: do nothing until release. Please copy here your `read` function. Maybe there is a small issue in it.

As you asked I merge this PR! Thank you very much for the updates! :)

Zaltora:
I found the problem. it is read to fast when i send return true. So i get alternate pressed/released state.
So it is ok. :)

kisvegabor:
Great! :)

PR #11, opened 2018-06-20
evdev and fbdev changes from 'master' into 'beta'

aenchevich:
Sync devel branch 'beta' to current 'master', which contains important fbdev- and evdev-related changes.


kisvegabor:
Thank you!

PR #13, opened 2018-06-21
Astyle

aenchevich:
Adding astyle config files with formatting rules mimicing project coding style and using them to auto-format code.


kisvegabor:
Thank you!

PR #15, opened 2018-06-26
Using hal_deinit to properly deinitialize the hardware abst. layer

aenchevich:


aenchevich:
A hal_deinit for the monitor device to properly deinitialize the SDL2 library when a QUIT event is detected.

PR #17, opened 2018-09-04
Properly redraw window when focused

carloantinarella:
With this added piece of code, the PC simulator window gets redrawn when it (re)gains focus or gets partially or totally uncovered by other windows. I tested it only on Linux, and it seems to work fine.


kisvegabor:
Merged, thank you!

PR #18, opened 2018-09-13
Adding #ifdef LV_CONF_INCLUDE_SIMPLE

aenchevich:


kisvegabor:
Hi,

`lv_drv_conf.h` should be included before `#if USE_....` because the header file contains `#define USE_....`

So instead of:
```c
#if USE_R61581

#ifdef LV_CONF_INCLUDE_SIMPLE
#include "lv_drv_conf.h"
#else
#include "../../lv_drv_conf.h"
#endif
```

this:
```c
#ifdef LV_CONF_INCLUDE_SIMPLE
#include "lv_drv_conf.h"
#else
#include "../../lv_drv_conf.h"
#endif

#if USE_R61581
```

Do you agree?

PR #25, opened 2018-09-25
Implement SDL encoder driver using mouse wheel.

poelstra:
Simple driver to use mouse wheel as encoder input for testing.


kisvegabor:
Awesome idea and simple implementation! 
Thank you very much! :+1:

PR #27, opened 2018-10-08
update ssd1306 with lvgl 5.2 (use VDB pixel based)

Zaltora:
Test with SPI3 + SH1106   ok

### DIFF
--- a/display/SSD1306.c
+++ b/display/SSD1306.c
@@ -17,6 +17,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include "lvgl/lv_core/lv_vdb.h"
+#include "lvgl/lv_core/lv_refr.h"
 
 /*********************
  *      DEFINES
@@ -94,6 +95,7 @@
 /**********************
  *  STATIC PROTOTYPES
  **********************/
+static void inline _rounder(lv_area_t *a);
 static int inline _load_frame_buffer(const ssd1306_t *dev, uint8_t* buf, uint8_t x1, uint8_t y1,  uint8_t x2, uint8_t y2);
 
 /**********************
@@ -126,13 +128,6 @@ static const ssd1306_t* _dev;
 /**********************
  *   GLOBAL FUNCTIONS
  **********************/
-void ssd1306_rounder(lv_area_t *a)
-{
-    /*Make the area bigger in y*/
-    a->y1 = a->y1 & ~(0x7);
-    a->y2 = a->y2 |  (0x7);
-}
-
 void ssd1306_vdb_wr(uint8_t * buf, lv_coord_t buf_w, lv_coord_t x, lv_coord_t y, lv_color_t color, lv_opa_t opa)
 {
     buf += buf_w * (y >> 3) + x;
@@ -238,6 +233,13 @@ int ssd1306_init(const ssd1306_t *dev)
     default:
         debug("Unsupported screen witdh");
         return -ENOTSUP;
+    }
+
+    //We send 8 pixel at once (page coordinate ). If VDB is not set correctly,
+    //we need round the y coordinate
+    if(LV_VDB_SIZE%8)
+    {
+        lv_refr_set_round_cb(_rounder);
     }
 
     switch (dev->protocol)
@@ -600,6 +602,13 @@ int ssd1306_start_scroll_hori_vert(const ssd1306_t *dev, bool way, uint8_t start
 /**********************
  *   STATIC FUNCTIONS
  **********************/
+static void inline _rounder(lv_area_t *a)
+{
+    /*Make the area bigger in y*/
+    a->y1 = a->y1 & ~(0x7);
+    a->y2 = a->y2 |  (0x7);
+}
+
 static int inline _load_frame_buffer(const ssd1306_t *dev, uint8_t* buf, uint8_t x1, uint8_t y1,  uint8_t x2, uint8_t y2)
 {
     uint8_t i;
--- a/display/SSD1306.c
+++ b/display/SSD1306.c
@@ -154,14 +154,14 @@ void ssd1306_flush(int32_t x1, int32_t y1, int32_t x2, int32_t y2, const lv_colo
         lv_flush_ready();
         return;
     }
+    /*Return if the screen is uninitialized*/
+    if(!_dev)
+    {
+        lv_flush_ready();
+        return;
+    }
 
-    /*Truncate the area to the screen*/
-    int32_t act_x1 = x1 < 0 ? 0 : x1;
-    int32_t act_y1 = y1 < 0 ? 0 : y1;
-    int32_t act_x2 = x2 > LV_HOR_RES - 1 ? LV_HOR_RES - 1 : x2;
-    int32_t act_y2 = y2 > LV_VER_RES - 1 ? LV_VER_RES - 1 : y2;
-
-    _load_frame_buffer(_dev, (uint8_t*)color_p, act_x1, act_y1, act_x2, act_y2);
+    _load_frame_buffer(_dev, (uint8_t*)color_p, x1, y1, x2, y2);
     lv_flush_ready();
 }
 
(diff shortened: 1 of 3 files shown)


Zaltora:
I got a problem. I need to fix color order.
I want set background  to black but i don't no where to set it.

Zaltora:
I found the correct setting:
```c
    //Set a style
    lv_style_copy(&style, &lv_style_transp);
    style.text.font = &lv_font_dejavu_10; 
    style.text.color.full = 1;
    style.body.main_color.full = 0;
    style.body.grad_color.full = 0;
    style.body.shadow.color.full = 0;
    style.body.border.color.full = 0;
    style.body.empty = 0;

    //Create main screen obj
    lv_obj_t * scr = lv_obj_create(NULL, NULL);
    lv_scr_load(scr);
    lv_obj_set_style(scr, &style);

    //Create a simple label
    label = lv_label_create(lv_scr_act(), NULL);
    lv_obj_set_style(label, &style);
```
A build-in style for monochrome display can be interesting for users. Wrong settings on a style and you got some strange behavior.
When i change `style.body.empty = 0;` to `1`, i got some glitch depending of VDB size.

Zaltora:
My settings:
```c
#define LV_HOR_RES          (128)
#define LV_VER_RES          (64)
#define LV_VDB_SIZE         (LV_VER_RES * LV_HOR_RES / 2) 
```

> If the bytes are mapped vertically on the display then it should be at least 8 * LV_HOR_RES 

I don't understand why. When i use a smaller VDB like `(LV_VER_RES * LV_HOR_RES / 10) `
, the display still work very well. I think the rounder function do the job.

---
The driver work well.
I got just a visual problem with wrong style setting for monochrome ( i think it is normal ?).
When i use `style.body.empty = 1;` and use a VDB size of  `(LV_VER_RES * LV_HOR_RES / 2) `,
Lvgl will write 2 time the VDB, one in top and one in bottom. 
So lvgl will write the label in top side like i want. A simple text.
For the bottom side, lvgl need to write nothing. I set no objects at all

But lvgl will send the same VDB  buffer in the screen bottom  with the same data as send for the top, result to get 2 time the same thing at the screen. One in the bottom, one in the top.
If you get smaller VDB, you will get a number of copy that is the same that the amount of VDB buffer to send to the screen for update it.

kisvegabor:
> I don't understand why. When i use a smaller VDB like (LV_VER_RES * LV_HOR_RES / 10)
, the display still work very well. I think the rounder function do the job.

Ye,s maybe it solves it somehow. But if you got strange things, remember to try other the VDB_SIZE :)

>  got just a visual problem with wrong style setting for monochrome ( i think it is normal ?).
When i use style.body.empty = 1; and use a VDB size of (LV_VER_RES * LV_HOR_RES / 2),

I got it! With `body.empty = 1` the object won't be drawn (only the border and/or shadow). Due to this, the previous data will remain in the VDB because there is nothing to overwrite it. So typically you shouldn't set `body.empty = 1` on the screen to have at least one object which will result in a solid background.

Zaltora:
Driver update ok for me :)

kisvegabor:
Looks great! :)

What do you think, would it be a good idea to register the `rounder` function in `ssd1306_init` because it's required for this display controller to work properly?

Zaltora:
Well i was thinking this too but if the user set a good VDB size, the rounder is not needed.
Maybe add a warning ? What you think ?

kisvegabor:
Rounder is needed with every VDB size. For example, if only 1 pixel changes LittelvGL still needs to redraw 8 pixels.

Zaltora:
OK rounder do the job. e.g:
`LV_VER_RES * LV_HOR_RES / 2 ` => rounder not needed
`LV_VER_RES * LV_HOR_RES / 5`  => rounder used.
Tested with and without the rounder. The difference is here !

When i set `LV_VER_RES * LV_HOR_RES / 10` , flush function is never called. Protection agains't too small VDB ? (use map and fill instead ?)

Zaltora:
> Rounder is needed with every VDB size. For example, if only 1 pixel changes LittelvGL still needs to redraw 8 pixels.

A, yeah forget about that. I will update again :)

kisvegabor:
Yes, `flush` is protected if the VDB was not big enough because of the rounder.
So the minimum is `LV_HOR_RES * 8` = 128 byte.
If the controller can be configured to map the pixel horizontally then `LV_HOR_RES` would be enough which is only 16 bytes. It's so few that already sounds funny :D

Zaltora:
How the minimum is determined ? Why a VDB less than 1024 byte can't work with this screen ?
If the controller was set horizontally, the minimum is not  `LV_VERT_RES * 8` ?

I don't understand well this part.

kisvegabor:
Explanations:
1. Let's say you have a 128x64 display where the bytes are mapped vertically
2. You have 8 x LV_HOR_RES = 8 x 128 = 1024 px VDB
3. A 55x42 area needs to be drawn (42 is already rounded to 8). 
4. LittelvGL calculates how many rows can be drawn into the VDB: 1024 (VDB SIZE) / 55 (area width) = 18,6
5. The number of rows is rounded down 18,6->16. So in this case, 55x16 areas can be refreshed at a time. (Check: 55*16 = 880. 880 < 1024 so it really fits into the VDB)
6. So the library will redraw the 55x48 area in 3 parts: 48x16 + 48x16 + 48x8. 

If the VDB would be 3xLV_HOR_RES and the full screen needs to be redrawn then: The the VDB can hold 3 lines. But you need at least 8 lines to properly refresh the display.

However,  if bytes were mapped horizontally you would need a rounder function in the X direction. So if your VDB is 1xLV_HOR_RES and the full-screen needs to be refreshed the VDB can hold 1 line and it's fine because the library rounds down only the X coordinate (128->128).

Zaltora:
If i understand well. VDB calculation is based with a complete row ? This computation can't be reduced to half a row for example ?
If VDB is 512 for example, the number become 9,3-> 8 .  What i missing ?
6- Are you sure for the number ? The area sum of theses 3 part is inferior the total area
I Need to understand better this VDB calculation part ^^.

How is checked this limitation ? It is hardcoded `8 x LV_HOR_RES `?
If we got a limitation, I think the function `lv_refr_set_round_cb`  need return an error if VDB is too small.

Ps: sry for late answer

kisvegabor:
> If i understand well. VDB calculation is based with a complete row ? This computation can't be reduced to half a row for example ?

Yes, it is based on the whole row and can't be reduced now. I think VDB can have a very small size now and with further optimization (i.e. half row support) we could gain only a few bytes of memory.

> If VDB is 512 for example, the number become 9,3-> 8 . What i missing ?

512/55 is really 9,3 and can work, but if the full screen needs to be redrawn not this small area. In this case, 512/128 = 4 is not ok.

> 6- Are you sure for the number ? The area sum of theses 3 part is inferior the total area

I really messed up the numbers a little bit. I edited the post to fix it.

> How is checked this limitation ? It is hardcoded 8 x LV_HOR_RES?
If we got a limitation, I think the function lv_refr_set_round_cb need return an error if VDB is too small.

It works like this:
1. Let's say we have a 512 px VDB and we redraw an 51x32 area.
2. 512/51 = 10 so we have a 51x10 area at the first step. 
3. The library needs to make it to 55x8. The rounder function rounds up the coordinates but we need to round them down. To do this the library starts to decrement the height until the rounded value greater than the original:
- 51x10 -> 51x16: wrong too big
- 51x9 -> 51x16: wrong, too big
- 51x8 -> ok
So the minimum size is not hardcoded but it is determined by the rounder function. 

> Ps: sry for late answer

No worries :)

Zaltora:
OK then :).
This PR is ok for me.

I just wonder how make more obvious when the VDB is to low. I updated the driver to be sure that the minimum required is here.

kisvegabor:
Great!
Please let me know if I can merge this PR.

Zaltora:
You can :)

kisvegabor:
Thank yu very much! :)

PR #32, opened 2018-10-27
fbdev: adding support for monochrome framebuffer

ederguidsa:
Currently, fbdev driver does not support monochrome framebuffers.
This pull request addresses this issue. I've tested on linux with ssd1307 frame buffer driver (https://github.com/torvalds/linux/blob/master/drivers/video/fbdev/ssd1307fb.c) and it worked perfectly.

### DIFF
--- a/display/fbdev.c
+++ b/display/fbdev.c
@@ -88,6 +88,11 @@ void fbdev_init(void)
 
 }
 
+void fbdev_exit(void)
+{
+    close(fbfd);
+}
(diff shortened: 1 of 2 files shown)


ederguidsa:
Additionally I've added a 'fbdev_exit' function in order to allow a clean exit, whenever it's needed.

kisvegabor:
Well done! Thank you very much!

Issue #35, opened 2018-11-08
Scale input received from evdev, if touchscreen resolution does not match display resolution


mheranco:
Hi,

the resolution of my touch input reveived from evdev does not match the resolution of my display.
With the following patch it would be possible to define a custom input resolution. Every position read from evdev would be scaled to the display resolution.

```diff --git a/b/indev/evdev.c b/a/indev/evdev.c
index 94027d8..e7346f7 100644
--- a/b/indev/evdev.c
+++ b/a/indev/evdev.c
@@ -90,21 +90,25 @@ bool evdev_read(lv_indev_data_t * data)
         }
     }

-
-    if(evdev_root_x < 0)
-        evdev_root_x = 0;
-    if(evdev_root_y < 0)
-        evdev_root_y = 0;
-    if(evdev_root_x >= LV_HOR_RES)
-        evdev_root_x = LV_HOR_RES - 1;
-    if(evdev_root_y >= LV_VER_RES)
-        evdev_root_y = LV_VER_RES - 1;
-
     /*Store the collected data*/
+#if SCALE_EVDEV
+    data->point.x = (evdev_root_x * LV_HOR_RES) / SCALE_EVDEV_HOR_RES;
+    data->point.y = (evdev_root_y * LV_VER_RES) / SCALE_EVDEV_VER_RES;
+#else
     data->point.x = evdev_root_x;
     data->point.y = evdev_root_y;
+#endif
     data->state = evdev_button;

+    if(data->point.x < 0)
+      data->point.x = 0;
+    if(data->point.y < 0)
+      data->point.y = 0;
+    if(data->point.x >= LV_HOR_RES)
+      data->point.x = LV_HOR_RES - 1;
+    if(data->point.y >= LV_VER_RES)
+      data->point.y = LV_VER_RES - 1;
+
     return false;
 }
```

```diff --git a/b/lv_drv_conf_templ.h b/a/lv_drv_conf_templ.h
index 449796c..2090830 100644
--- a/b/lv_drv_conf_templ.h
+++ b/a/lv_drv_conf_templ.h
@@ -207,6 +207,11 @@
 #define USE_EVDEV    0
 #if USE_EVDEV
 #define EVDEV_NAME   "/dev/input/event0"        /*You can use the "evtest" Linux tool to get the list of devices and test them*/
+#define SCALE_EVDEV  0                          /* Scale input, e.g. if touchscreen resolution does not match display resolution */
+#if SCALE_EVDEV
[...]

PR #36, opened 2018-11-08
evdev: scale input if touchscreen resolution does not match display res

mheranco:
See Issue: #35


kisvegabor:
Merged, thank you!

PR #40, opened 2018-11-21
#ifndef guards for all USE_<device> defines in lv_drv_conf

aenchevich:


kisvegabor:
Thank you! It will make easier to work with defines in makefile!

PR #41, opened 2018-11-21
rename encoder to mousewheel - part 2

aenchevich:
adding a couple of lines which were missing from the original commit "rename evdev to mousewheel "


kisvegabor:
Thank you!

PR #45, opened 2018-12-06
Added touch screen device file setting function of evdev.

chaos4284:
Added evdev_set_file() function.


kisvegabor:
Awesome! Thank you!

PR #47, opened 2019-01-03
Patches to support Emscripten in `monitor` driver

embeddedt:


kisvegabor:
I noticed two things:
1. As mentioned above, once you have `MONITOR_EMSCRIPTEN` it would be cleaner to use it later instead of `__EMSCRIPTEN__`
2. It seems you if we are using Emscripten you enable `MONITOR_APPLE`. It's not so clean. Is there any disadvantages to use `MONITOR_APPLE || MONITOR_EMSCRIPTEN`?

embeddedt:
@kisvegabor Both of those are good points. I adjusted the `#ifdef`s.

kisvegabor:
Perfect! Thank you!

kisvegabor:
I usually delete feature branches on merge. Is it not a problem for you?

embeddedt:
@kisvegabor No problem. There is no point keeping an outdated feature-specific branch around forever.

PR #54, opened 2019-05-28
Added event handling for touchpanel events

bertthomas:


bertthomas:
Short answer: I don’t. It simply just responds to X and Y positions, regardless of what finger. I assume that lvgl does not yet have support for actions with multiple fingers such as zooming on a cell phone works. The result of the current change is that the “mouse pointer” moves to the spot where the last finger touched the screen. To be honest, this was not by design but just a “way to get it to work”.  Despite that I think that the patch might be useful for other users as well even though it could be improved  as otherwise LVGL doesn’t respond to any MT events.

embeddedt:
@bertthomas 

> The result of the current change is that the “mouse pointer” moves to the spot where the last finger touched the screen.

I didn't realize that it behaved in this way. That would be the expected behavior on our end, so no problems there.

PR #55, opened 2019-06-27
libinput_read should return touch coordinates in pixel not in mm

mheranco:
At the moment `libinput_read` returns the touch coordinates in `mm`, not in pixel as littlevgl expects. The `libinput_event_touch_get_x_transformed` simply maps the touch coordinate to the given width parameter. Changing this parameter from max screen size in `mm` to max screen size in pixel fixes the problem.

### DIFF
--- a/indev/libinput.c
+++ b/indev/libinput.c
@@ -136,8 +136,8 @@ bool libinput_read(lv_indev_data_t * data)
       case LIBINPUT_EVENT_TOUCH_MOTION:
       case LIBINPUT_EVENT_TOUCH_DOWN:
         touch_event = libinput_event_get_touch_event(event);
-        most_recent_touch_point.x = libinput_event_touch_get_x_transformed(touch_event, LV_HOR_SIZE);
-        most_recent_touch_point.y = libinput_event_touch_get_y_transformed(touch_event, LV_VER_SIZE);
+        most_recent_touch_point.x = libinput_event_touch_get_x_transformed(touch_event, LV_HOR_RES);
+        most_recent_touch_point.y = libinput_event_touch_get_y_transformed(touch_event, LV_VER_RES);
         libinput_button = LV_INDEV_STATE_PR;
         break;
       case LIBINPUT_EVENT_TOUCH_UP:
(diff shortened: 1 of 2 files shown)


kisvegabor:
Hi,

By reading the documentation [here](https://wayland.freedesktop.org/libinput/doc/1.2.1/group__event__touch.html#gadd04ccc3a7b4564e5aca3c8592f090fe), it's not clear what should be the `width` parameter.

@leiflm Can you comment on this?
(He added this driver)

mheranco:
Hi, 

yes the documentation is unclear.

I looked into the source code:
[libinput_event_touch_get_x_transformed](https://gitlab.freedesktop.org/libinput/libinput/blob/master/src/libinput.c#L833) calls [evdev_device_transform_x](https://gitlab.freedesktop.org/libinput/libinput/blob/master/src/evdev.c#L334) which calls [scale_axis](https://gitlab.freedesktop.org/libinput/libinput/blob/master/src/evdev.c#L327) which simply maps the value.

leiflm:
I believe @mheranco is right about this and hist patch should be applied.

In the current way it targets pixel agnostic rendering environments, using `mm` as unit of measure.

AFAIK Littlevgl is heavily based on pixels as unit of measure though.

Once littlevgl converts to `mm` instead of `px` this could be a compile time switch.

embeddedt:
> Once littlevgl converts to `mm`

I don't think there are plans to do that at present. We are targetting embedded systems with a constant DPI.

leiflm:
> I don't think there are plans to do that at present. We are targetting embedded systems with a constant DPI.

On embedded systems, you sometimes see touchscreens mounted onto displays. On Linux, specifically, there are [properties to specify your touchscreen's geometry](https://www.kernel.org/doc/Documentation/devicetree/bindings/input/touchscreen/touchscreen.txt). What you can get from there are coordinates in `mm`, that will always be correct, no matter the resolution of the display mounted beneath it.

So when your product's displays gets a DPI upgrade, that wouldn't affect your drawing code and labels wouldn't become tiny all of a sudden.

But, as I wrote, I don't think the time has come for DPI independent stuff in littlevgl yet.

embeddedt:
You can actually do DPI independent things in LittlevGL already, by simply handling all of your sizes based on `LV_DPI` multiplied and/or divided by some constant.

leiflm:
> You can actually do DPI independent things in LittlevGL already

Sorry for my uninformed statement and thank you for the hint.

embeddedt:
@leiflm No worries! :smiley: I just wanted to clarify that DPI isn't completely unsupported.

PR #58, opened 2019-07-11
poll does not set errno, when it returns 0

mheranco:
Poll only sets `errno`, if return value is `-1`.


embeddedt:
Thanks!

A second option would be to set `errno = 0` before calling `poll`, however, I think that could result in a stream of unwanted "Success" messages.

PR #60, opened 2019-07-13
uc1610 display controller support

manalpham:
Hi,
to support uc1610 4 gray display controller (tested with EA DOGXL160 160x104 screen)


kisvegabor:
Awesome, thank you!

PR #63, opened 2019-07-21
Sharp memory in pixels displays support

manalpham:
Hi,
to support Sharp MIP monochrome displays.
Tested with LS027B7DH01 (400x240 pixels).

### DIFF
--- a/display/SHARP_MIP.c
+++ b/display/SHARP_MIP.c
@@ -127,14 +127,6 @@ void sharp_mip_flush_cb(lv_disp_drv_t * disp_drv, const lv_area_t * area, lv_col
   /* Set frame header in VDB */
   buf[0] = SHARP_MIP_HEADER         |
            SHARP_MIP_UPDATE_RAM_FLAG;
-#if SHARP_MIP_SOFT_COM_INVERSION
-  if (com_output_state) {
-    com_output_state = false;
-  } else {
-    buf[0] |= SHARP_MIP_COM_INVERSION_FLAG;
-    com_output_state = true;
-  }
-#endif
 
   /* Write the frame on display memory */
   LV_DRV_DISP_SPI_CS(1);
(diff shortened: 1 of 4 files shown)


manalpham:
703c24b & 711dda8 commits are not relevent (do-undo on master), sorry for the noise

embeddedt:
This is the only commit you want merged? https://github.com/littlevgl/lv_drivers/pull/63/commits/0b9315274cfd7bb2a29b8e1dd4f170a2a31a2fe3

manalpham:
Yes it is. 0b93152

embeddedt:
@manalpham The driver looks fine to me.

I have a question about the COM inversion. Is it really that reliable to update it during the flush callback? That will only run when the display needs to be refreshed - if the screen only changes on a touch it wouldn't be inverted until the next touch on the screen.

manalpham:
yes, i agree, flush callback isn't the right place for com inversion.
I purpose this 9b8d27e to let user invert com periodically (0.5 - 10 Hz datasheet) by calling `sharp_mip_com_inversion()` function.
Unfortunately, i can't test soft com inversion because my board is wired with EXTMODE high so the com invertion is hard (with dedicated EXTCOMIN pin)

embeddedt:
I think that's an acceptable tradeoff.

@kisvegabor If you don't have any objections this can be merged by cherry-picking (in order) https://github.com/littlevgl/lv_drivers/pull/63/commits/0b9315274cfd7bb2a29b8e1dd4f170a2a31a2fe3, https://github.com/littlevgl/lv_drivers/pull/63/commits/9b8d27e5a8ae1e7d19abac0ad596e51fc91108a7, and https://github.com/littlevgl/lv_drivers/pull/63/commits/fc5074e66978564ba209a6dae6c37f2f9be98544.

kisvegabor:
@embeddedt 
I agree, it looks good to me too. Only one minor thing: in other drivers the there is no `_cb` postfix in the name of API functions. Could we change them? 
E.g. `sharp_mip_flush_cb` :arrow_right:  `sharp_mip_flush`

kisvegabor:
I've fixed it and merged.

PR #64, opened 2019-07-22
Fix libinput_read function args.

summershrimp:
`lv_indev_drv_t.read_cb` has changed. So we need to adapt it.


kisvegabor:
Thank you!

PR #69, opened 2019-08-01
Add PlatformIO library config

puzrin:
Tested locally, seems to work. Will retest again after merge.

FYI: to work with PIO deps locally, remove those from dependencies & put clone to `./lib` folder of project.


kisvegabor:
I've fixed a typo and added a few more keywords.

PR #71, opened 2019-08-02
Change libinput.h to libinput_drv.h

summershrimp:
Since the indrv.mk add the path to include directories,
compiler cannot determine which libinput.h to read only by using
`#include <libinput.h>` and `#include "libinput.h"`


indev/libinput.c
```diff
@@ -6,7 +6,7 @@
 /*********************
  *      INCLUDES
  *********************/
-#include "libinput.h"
```
summershrimp: confilct with line 19  `#include <libinput.h>`

embeddedt:
Good point. Thanks!

PR #72, opened 2019-08-02
monitor: add fallbacks for hor/ver values, close #70

### LINKED ISSUE #70: Defaults for MONITOR_HOR_RES & MONITOR_VER_RES ?
https://github.com/littlevgl/lv_platformio/blob/master/platformio.ini#L26-L27

Now need to define MONITOR_HOR_RES & MONITOR_VER_RES in every platform.ini. Not a problem, but in 99% looks like useless garbage.

May be worth add such fallback defines near MONITOR_ZOOM ?

https://github.com/littlevgl/lv_drivers/blob/master/display/monitor.c#L29-L31

puzrin:


kisvegabor:
Thank you!

PR #73, opened 2019-08-02
pio: add `version` property to force crawler scan git tags

puzrin:
Also need git tag `v6.0.2`


kisvegabor:
Tagged and released.

PR #74, opened 2019-08-10
Add support for FreeBSD framebuffer

thelunatic:
This commit adds frame buffer driver support for FreeBSD and it has been tested
with a sample app on freebsd-arm image Beagleboneblack 12-STABLE


embeddedt:
Would it be possible to add some conditional logic to the Linux framebuffer driver instead? It would prevent duplication of common logic - since the only difference is really how the framebuffer pointer is retrieved.

thelunatic:
Some of the include headers would be different as well. Maybe some code guards can keep them separate like USE_BSD_FBDEV. Mostly the init function is different as you noted, but including it in the fbdev.c with code guards might not be a very clean approach I guess. What do you suggest?

embeddedt:
Even the init function is quite similar. Both platforms `open` the framebuffer, run an `ioctl` to get information, and then `mmap` the opened file into memory.

So from what I can tell, only two code guards are necessary (one in the init function, and one for the includes).

Is that easy to do? If not we can just keep two separate drivers, but we'd have to remember to sync them for common stuff.

thelunatic:
I understand your concern about adding an extra driver where a few code guards could make it simpler to maintain. I'll have a look into it and try running it on the board. I'll report here when I get it working or if I can't get it working.
Thanks for reviewing. :)

thelunatic:
I had a more detailed look into the linux framebuffer and i see the use of some data structures like fb_var_screeninfo. The structures in freeBSD are very different (both in name and organizing)[1]. I think there isn't any other option but to maintain a separate driver for FBSD.
[1]: https://github.com/freebsd/freebsd/blob/master/sys/sys/fbio.h

embeddedt:
> The structures in freeBSD are very different (both in name and organizing)[1]. I think there isn't any other option but to maintain a separate driver for FBSD.

We store all of the necessary information here:

https://github.com/littlevgl/lv_drivers/blob/5ef341c03f2eed1fde6fa66076885e4e0cb7c2fb/display/bsd_fbdev.c#L47-L52

Shouldn't it be possible to just do:

```c
#ifdef USE_BSD_FBDEV
/* get BSD framebuffer information and store it in variables */
#else 
/* get Linux framebuffer information and store it in variables */
#endif
```

thelunatic:
Hi,
I have added the freebsd support in fbdev.c. Please have a  look and squash and merge it if you're OK with the commit.

Thanks

display/fbdev.c
```diff
@@ -49,7 +78,60 @@ static int fbfd = 0;
 /**********************
  *   GLOBAL FUNCTIONS
  **********************/
+#ifdef USE_BSD_FBDEV
+void fbdev_init(void)
```
embeddedt: Can this be consolidated further? Mostly lines 97 to 114 need to be split.

thelunatic:
Hi
I have made the changes accordingly, please have a look.

kisvegabor:
It looks good to me, thank you very much!

@embeddedt ?

embeddedt:
Since it's not a breaking change and it is not part of core LittlevGL I will just merge it straight to `master`.

PR #79, opened 2019-09-16
fbdev.c: fix using sec-screen-buffer crash

lgl88911:
Signed-off-by: Frank Li <lgl88911@163.com>

### DIFF
--- a/display/fbdev.c
+++ b/display/fbdev.c
@@ -129,7 +129,7 @@ void fbdev_init(void)
     printf("%dx%d, %dbpp\n", vinfo.xres, vinfo.yres, vinfo.bits_per_pixel);
 
     // Figure out the size of the screen in bytes
-    screensize =  finfo.line_length * vinfo.yres;
+    screensize =  finfo.smem_len; //finfo.line_length * vinfo.yres;    
 
     // Map the device to memory
     fbp = (char *)mmap(0, screensize, PROT_READ | PROT_WRITE, MAP_SHARED, fbfd, 0);


embeddedt:
(cc @kisvegabor)

I am not sure of the exact scope of this change. Can you describe the issue you experienced in more detail? Why isn't `line_length * yres` equivalent to the size of your screen?

lgl88911:
> (cc @kisvegabor)
> 
> I am not sure of the exact scope of this change. Can you describe the issue you experienced in more detail? Why isn't `line_length * yres` equivalent to the size of your screen?

In my platform, 
printf("%d(%d)(%d)x%d(%d)(%d), %dbpp linebyte: %d\n", vinfo.xres, vinfo.xoffset, vinfo.xres_virtual, vinfo.yres, vinfo.yoffset, vinfo.yres_virtual, vinfo.bits_per_pixel, finfo.line_length);
show blow:
1920(0)(1920)x1080(2160)(3240), 32bpp linebyte: 7680
current display from offset 2160, 
If only map 0~finfo.line_length * vinfo.yres,  during access "(act_x1 + vinfo.xoffset) + (y + vinfo.yoffset) * finfo.line_length / 4;" will out of memory cause crash。
So need map all framebuffer by using smem_len

kisvegabor:
So the problem is that the offsets are not taken into account?

lgl88911:
Yes, If framebuffer is double buffer and display screen point to secend buffer. Current code only map first buffer, but access secend buffer

kisvegabor:
I see. `smem_len` seems also [standard field](https://www.kernel.org/doc/Documentation/fb/api.txt) in `fb_fix_screeninfo`. 

Merged, thank you!

PR #83, opened 2019-10-18
Updated touch input templates

JojoS62:
- updated signature for read functions
- fixed wrong readings with XPT2046 driver, was causing a jumping cursor


kisvegabor:
Than kyou for the fix!

PR #87, opened 2019-11-21
Correct X, Y axis not inverting when EVDEV_XXX_MIN > EVDEV_XXX_MAX

criscola:


criscola:
Ok so I should just replace the macro with ``lv_disp_get_xxx_res(drv->disp)`` in the four map functions so it can work when there are multiple displays?

embeddedt:
Yes. Essentially, revert that portion of your diff. You can see how it should be [here](https://github.com/littlevgl/lv_drivers/pull/87/files) (in the red part).

embeddedt:
Looks fine now. Thanks!

criscola:
Thanks to you, this was my first pull-request yay!

PR #89, opened 2020-01-11
Use truncated width in fbdev.c

### LINKED ISSUE #88: fbdev.c: Re-use of previously confined area can result in out-of-range memcpy
The function fbdev_flush nicely confines the set area if it exceeds the screen size.

But in the line https://github.com/littlevgl/lv_drivers/blob/master/display/fbdev.c#L175 the original unconfined area value is used for width calculation which can result in memory corruption.

I don't have a monochrome display but line https://github.com/littlevgl/lv_drivers/blob/master/display/fbdev.c#L225 may also be affected by this problem.

embeddedt:
May fix #88.


kisvegabor:
Thank you!

PR #93, opened 2020-03-15
fix scaling, when EVDEV_CALIBRATE is false and EVDEV_SCALE true

### LINKED ISSUE #92: Touchscreen Problem after upgrading lv_drivers folder
After updating the lv_drivers folder for our application, the X/Y coordinates being reported to our application are not correct. Looks like the problem is in evdev.c and is related to the check in commited on Nov 24, 2019.

In our case, the data->point.x/y calculated in the #if EVDEV_SCALE section is being overwritten by the #else portion of the #if EV_CALIBRATE section (lines 189-199). 

I am not sure what case the Nov 2019 checkin was addressing but looks like the code will now always override the x/y values calculated using the scaling code. So I'm not sure if that check in should just be reverted or if it needs modified in some manner.

knicklux:
This closes #92. data->point.x = evdev_root_x was always overriding the scaled x-coordinate, because if EVDEV_SCALE is true, EVDEV_CALIBRATE would be false and thus the non-scaled values would still be used.

### DIFF
--- a/indev/evdev.c
+++ b/indev/evdev.c
@@ -189,10 +189,9 @@ bool evdev_read(lv_indev_drv_t * drv, lv_indev_data_t * data)
 #if EVDEV_SCALE
     data->point.x = map(evdev_root_x, 0, EVDEV_SCALE_HOR_RES, 0, lv_disp_get_hor_res(drv->disp));
     data->point.y = map(evdev_root_y, 0, EVDEV_SCALE_VER_RES, 0, lv_disp_get_ver_res(drv->disp));
-#endif
-#if EVDEV_CALIBRATE
-	data->point.x = map(evdev_root_x, EVDEV_HOR_MIN, EVDEV_HOR_MAX, 0, lv_disp_get_hor_res(drv->disp));
-	data->point.y = map(evdev_root_y, EVDEV_VER_MIN, EVDEV_VER_MAX, 0, lv_disp_get_ver_res(drv->disp));
+#elif EVDEV_CALIBRATE
+    data->point.x = map(evdev_root_x, EVDEV_HOR_MIN, EVDEV_HOR_MAX, 0, lv_disp_get_hor_res(drv->disp));
+    data->point.y = map(evdev_root_y, EVDEV_VER_MIN, EVDEV_VER_MAX, 0, lv_disp_get_ver_res(drv->disp));
 #else
     data->point.x = evdev_root_x;
     data->point.y = evdev_root_y;


embeddedt:
Thank you!

@AnneBruner Can you verify that this fixes your problem?

AnneBruner:
Yes that fixes my problem. Thx much!

embeddedt:
Glad to hear it!

mibcat:
Hi !
Unfortunately this commit reverts the pull-request [87](https://github.com/littlevgl/lv_drivers/pull/87) and the axis invertion does not work if **EVDEV_SCALE** = 1 and **EVDEV_CALIBRATE** = 1

kisvegabor:
@mibcat 
Oh, can you send a PR to fix it based on the new code?

mibcat:
Hi !
Obviously I misinterpret the meaning of both parameters: looking deeper at the software I think they should **not** be active at the same time, right ?
If so there is no fix needed (inversion works with EVDEV_SCALE = **0** and EVDEV_CALIBRATE = **1**.
But maybe a hint (or better a check) should be implemented that the parameters are used mutually exclusive.

mibcat:
Or another approach:

provide only **one** parameter for scaling and set the default values
```c
EVDEV_HOR_MIN 0
EVDEV_HOR_MAX 4096
EVDEV_VER_MIN 0
EVDEV_VER_MAX 4096
```
and in the comment a description how to scale and invert the axis ...

kisvegabor:
@mibcat 
What's the point of scale? Calibrate can do the same and even more, right?

mibcat:
yes, `EVDEV_CALIBRATE` is the more flexible one
`EVDEV_SCALE` is a _special case_ of `EVDEV_CALIBRATE` with the above mentioned values

kisvegabor:
Then fo simplicity I think we can keep only calibrate.

mibcat:
yeah - sometimes less is more ;-)

kisvegabor:
Great! :slightly_smiling_face: 

If you already have an environment to test it, can you send a PR?

mibcat:
Sure. But currently I'm on branch **release/v6** (7ac9388aed612c0cff91de92e3004cdfd9679629)

I've tried yesterday to update to **v7** - but there are a lot of changes in handling themes and styles which has an dramatically impact to my "modal message box" implementation ;-)

Because eliminating `EVDEV_SCALE` is also code breaking the PR should _not_ be done in v6.

==> therefore the PR could take some time ...

mibcat:
ahh, sorry - wrong repo (lvgl)

but to which branch should I do the PR for **lv_drivers** ? **dev-7.0** ?

kisvegabor:
It can go to `master`.

mibcat:
PR created (#99)

kisvegabor:
Merged, thank you!

PR #94, opened 2020-03-21
fbdev.c: fix BSD screensize error

thelunatic:
Commit ca3eed87bab2d3d341659d8512c850ef29a179c1 broke the BSD build. This patch is to adapt to that change.


embeddedt:
Thanks!

PR #99, opened 2020-05-22
EVDEV_SCALE removed which can be handled as a special case by EVDEV_CALIBRATE

mibcat:
## reason for changing
`EVDEV_CALIBRATE` provides the more flexible functions (scale & invert the axis) than `EVDEV_SCALE`. Also the existence of these two parameters caused issues and PRs in the past.

## changes
**EVDEV_SCALE** was removed and the settings for calibration was set to the default values of **EVDEV_SCALE**. 

### unwanted changes in functionality
If a project is using `EVDEV_SCALE` the scaling will simply not working without any notice to the user during build. But changing to `EVDEV_CALIBRATE` will restore the functionality.

## calibration hints
The previously used settings for `EVDEV_CALIBRATE` corresponds to typical 3.5 LCD touch panel used for a Raspberry PI operating in 180 degree screen rotation (`EVDEV_HOR_MIN` > `EVDEV_HOR_MAX`):

```c
#define EVDEV_HOR_MIN      3800
#define EVDEV_HOR_MAX       200
#define EVDEV_VER_MIN       200
#define EVDEV_VER_MAX      3800
```

### DIFF
--- a/indev/evdev.c
+++ b/indev/evdev.c
@@ -186,10 +186,7 @@ bool evdev_read(lv_indev_drv_t * drv, lv_indev_data_t * data)
         return false;
     /*Store the collected data*/
 
-#if EVDEV_SCALE
-    data->point.x = map(evdev_root_x, 0, EVDEV_SCALE_HOR_RES, 0, lv_disp_get_hor_res(drv->disp));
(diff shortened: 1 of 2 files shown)


kisvegabor:
Thak you!

PR #100, opened 2020-06-03
Add support for BSD evdev

thelunatic:


display/fbdev.c
```diff
@@ -108,7 +108,7 @@ void fbdev_init(void)
 
     vinfo.xres = (unsigned) fb.fb_width;
     vinfo.yres = (unsigned) fb.fb_height;
-    vinfo.bits_per_pixel = fb.fb_depth + 8;
+    vinfo.bits_per_pixel = fb.fb_depth;
```
kisvegabor, line 111: Why was it required?

kisvegabor:
Thank you! I had only one question.

display/fbdev.c
thelunatic: I think this was required with previous versions of lv_drivers, to make it 24 bits per pixel, it wasn't working without this. This is certainly a hack that's no more required and it's causing problems with the alignment of widgets. :)

kisvegabor: I see, thank you!

kisvegabor:
All clear, merged.

Thank you!

PR #104, opened 2020-07-30
Use wildcard function instead of find.

satirebird:
This fix uses the wildcard function instead of find because it allows to use other subdirectories for building.


kisvegabor:
Thank you!

PR #105, opened 2020-09-05
Fix wildcard matching in lv_drivers.mk

DolphinChips:
233860e introduced wildcard matching for sources in lv_drivers.mk,
but also introduced a typo, so gtkdrv/ directory was ignored and
non-existent gtk/ was searched. This resulted in linking error for
all programs using GTK as their display driver because gtkdrv/gtkdrv.c
wasn't compiled, which means that important symbols like gtkdrv_init
and gtkdrv_flush_cb weren't created.


kisvegabor:
Thanks, the make file was really outdated. But the directory has `gtkdrv` name, not `gtkdev`.

DolphinChips:
> 
> 
> Thanks, the make file was really outdated. But the directory has `gtkdrv` name, not `gtkdev`.

Fixed.

embeddedt:
Looks fine; thanks for the fix.

PR #110, opened 2020-10-16
display: drm: fix warnings

superna9999:
This fixes the warnings reported at https://forum.lvgl.io/t/drm-driver-is-slow/3479

Signed-off-by: Neil Armstrong <narmstrong@baylibre.com>


kisvegabor:
Thank you!

PR #114, opened 2020-11-15
Windows should not overlap when MONITOR_ZOOM > 1 and MONITOR_DUAL = 1 

mtauraso:
Quick fix to overlapping SDL windows when MONITOR_DUAL enabled and MONITOR_ZOOM greater than 1.


embeddedt:
Looks good; thank you for the fix!

PR #117, opened 2021-02-24
New native Windows driver (win32drv)

MouriNaruto:
## Differences from the old native Windows driver (win_drv)

- Much higher performance than the old native Windows driver and the SDL driver.
- Only depends on Win32 API and C Runtime.
- Written in pure C, so you can use it on x86, x64, ARM and ARM64 Windows.
- Add support for Per-monitor DPI Aware.
- Add support for keyboard and mouse wheel.
- Customize window icon support.

## Benchmark

The SDL driver
image

The new native Windows driver
image

### DIFF
--- a/win32drv/win32drv.h
+++ b/win32drv/win32drv.h
@@ -1,0 +1,60 @@
+﻿/**
+ * @file win32drv.h
+ *
+ */
+
+#ifndef LV_WIN32DRV_H
+#define LV_WIN32DRV_H
+
+/*********************
+ *      INCLUDES
+ *********************/
+
+#include <Windows.h>
+
+#if _MSC_VER >= 1200
+ // Disable compilation warnings.
+#pragma warning(push)
+// nonstandard extension used : bit field types other than int
+#pragma warning(disable:4214)
+// 'conversion' conversion from 'type1' to 'type2', possible loss of data
+#pragma warning(disable:4244)
+#endif
+
+#ifdef LV_LVGL_H_INCLUDE_SIMPLE
+#include "lvgl.h"
+#else
+#include "lvgl/lvgl.h"
+#endif
+
+#if _MSC_VER >= 1200
+// Restore compilation warnings.
+#pragma warning(pop)
+#endif
+
+/*********************
+ *      DEFINES
+ *********************/
+
+/**********************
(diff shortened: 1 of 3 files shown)


win32drv/win32drv.h
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+#ifndef LV_WIN32DRV_H
+#define LV_WIN32DRV_H
+
+/*********************
+ *      INCLUDES
+ *********************/
+
+#include <Windows.h>
```
embeddedt, line 13: I think this include needs to be gated behind a config option in `lv_drv_conf.h`, otherwise other users will get compilation errors if they build all files in this repo.

embeddedt (commented):
Looks good; just one comment.

kisvegabor:
Unbelievable!

 But what is the trick? :slightly_smiling_face:  It seems you just copy the image rendered by LVGL into your buffer: 
`memcpy(g_pixel_buffer, color_p, g_pixel_buffer_size);`

It's what most of the drivers do.

win32drv/win32drv.h
MouriNaruto: Got it. I will add it.

MouriNaruto:
> Unbelievable!
> 
> But what is the trick? 🙂 It seems you just copy the image rendered by LVGL into your buffer:
> `memcpy(g_pixel_buffer, color_p, g_pixel_buffer_size);`
> 
> It's what most of the drivers do.

First, I create the Windows GDI bitmap object *only once* in this implementation and get the memory address for writing the pixels. So, I can reuse the bitmap object. It will reduce most of the context switching overhead.

Second, I use `rounder_cb` to make copy the LVGL's pixel buffer to the bitmap object's pixels address easier.

In general, I had tried my best to reduce the memory copy operations. But I can't use my zero memory copy operation implementation for the universal driver because I can't limit the user only use the 32bpp color depth. (But in the repository for Windows desktop application, I can do that.)

win32drv/win32drv.h
MouriNaruto: Ok, I have added it.

kisvegabor:
Thanks! I'm surprised that memory copy is so expensive even on PC.

PR #118, opened 2021-03-14
Add Wayland support

WallaceIT:
Add Wayland display and input drivers, with support for keyboard, mouse and touchscreen.
Current implementation only supports `wl_shell` shell with no decorations.

Tested on Fedora 33.

### DIFF
--- a/lv_drivers.mk
+++ b/lv_drivers.mk
@@ -1,6 +1,7 @@
 LV_DRIVERS_DIR_NAME ?= lv_drivers
 
 CSRCS += $(wildcard $(LVGL_DIR)/$(LV_DRIVERS_DIR_NAME)/*.c)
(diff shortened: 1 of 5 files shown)


kisvegabor:
I naively tested it on Linux Mint Cinnamon 19 but it failed with 
```
lv_drivers/wayland/wayland.c:235: wayland_init: Assertion `application.display' failed.
``` 

I suppose it's because Wayland and X can't run together at all.

Anyway, I'm sure you tested it correctly so I merge it. Thank you!

WallaceIT:
AFAIK, Cinnamon runs only on X server and thus your system should not support Wayland clients - so the error you are obtaining makes sense.

Many thanks to you for your work!

kisvegabor:
Alright, thanks for the confirmation.

PR #119, opened 2021-03-16
Fix "clicking" on MT touchscreens.

luka177:
In linux kernel multi touch touchscreen drivers may not provide EV_KEYs events, and provide only ABS_MT_TRACKING_ID. This is very basic implementation, it is still single touch, but it allows to use those touchscreens.


embeddedt:
Thanks. This is probably all that's needed since LVGL doesn't have the concept of multi-touch at the moment.

PR #120, opened 2021-03-17
Add next/prev keys to SDL keyboard

boraozgen:
I noticed that the next/prev keys are missing from the SDL keyboard driver. I used the same keys from win32drv.


kisvegabor:
By default `tab` is mapped to "next" but with Page up/down we have "prev" too.
Thanks!

PR #121, opened 2021-03-17
Add initial CMake file

boraozgen:
Added a primitive CMake file to build a library. Actually globbing is not ideal (see [here](https://stackoverflow.com/questions/1027247/is-it-better-to-specify-source-files-with-glob-or-each-file-individually-in-cmak/1060061) but it can be improved later by swapping with explicit source filenames.

I'm doing this to build the SDL simulator in Windows. I'll detail it on its own PR.

Also fixes a wrong include path.


kisvegabor:
Thanks!

PR #126, opened 2021-05-11
Fix flush res, add screen shot.

turoksama:
As mentioned, what do you think?

### DIFF
--- a/display/monitor.c
+++ b/display/monitor.c
@@ -13,6 +13,7 @@
 #  define MONITOR_SDL_INCLUDE_PATH <SDL2/SDL.h>
 #endif
 
+#include <stdio.h>
 #include <stdlib.h>
 #include <stdbool.h>
 #include <string.h>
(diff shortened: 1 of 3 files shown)


embeddedt (commented):
IMO, it'd be better to add a native "screenshot_flush" function within LVGL that saves the screen output to a BMP file. In v8 I think the filesystem API is complete enough that we could have it as a feature within LVGL itself, and it should work regardless of the underlying filesystem.

kisvegabor:
I agree with @embeddedt. 

We could do something like this:
```c
void save_screenshot(lv_obj_t * obj, const char*  path)
{
   lv_refr_now(lv_obj_get_disp(obj));  //To redraws pending invlid areas
  
   //Set a new flush_cb in disp_drv
   ...
  
   //Redraw with the new flush_cb
   lv_obj_invalidate(obj); 
   lv_refr_now(lv_obj_get_disp(obj)); 
  
  //Restore the original flush_cb
  ...
}
```

This way not only a screen but any object can be saved. What do you think?

turoksama:
Wonderful!That will be full-featured lib besides PC more than a tweaks.
I'll close this pull and try in V8.

kisvegabor:
Great! :heart:

PR #135, opened 2021-07-07
Add input drivers to the default group for the new native Windows driver (win32drv)

MouriNaruto:
For solving "Still no keyboard and mouse." mentioned in https://github.com/lvgl/lv_sim_codeblocks_win/issues/16#issuecomment-875141993.

### DIFF
--- a/win32drv/win32drv.c
+++ b/win32drv/win32drv.c
@@ -125,15 +125,15 @@ static void lv_win32_display_driver_rounder_callback(
     lv_disp_drv_t* disp_drv,
     lv_area_t* area);
 
-static void lv_win32_mouse_driver_read_callback(
+static void lv_win32_pointer_driver_read_callback(
     lv_indev_drv_t* indev_drv,
     lv_indev_data_t* data);
 
-static void lv_win32_keyboard_driver_read_callback(
+static void lv_win32_keypad_driver_read_callback(
     lv_indev_drv_t* indev_drv,
     lv_indev_data_t* data);
 
-static void lv_win32_mousewheel_driver_read_callback(
+static void lv_win32_encoder_driver_read_callback(
     lv_indev_drv_t* indev_drv,
     lv_indev_data_t* data);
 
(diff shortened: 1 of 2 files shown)


embeddedt:
Interesting… I didn’t know pointer devices needed to be in a group?

MouriNaruto:
@embeddedt I have just added it like the implementation in lv_port_windows. (I thought maybe it can help me to enumerate input devices if someone needs.)

kisvegabor:
@embeddedt 
> Interesting… I didn’t know pointer devices needed to be in a group?

You need the assign the input device to a group to know what group should be affected by the inout device.


@MouriNaruto 
Shouldn't we add an `#if USE_WIN32DRV_DEFAULT_GROUP` to enable/disable these features?

MouriNaruto:
@kisvegabor 

> Shouldn't we add an `#if USE_WIN32DRV_DEFAULT_GROUP` to enable/disable these features?

I think "The keyboard and mousewheel input device hadn't been added to the default group" is more like the bug for the original win32drv implementation because win32drv hasn't export the C symbols for accessing input devices' lv_indev_drv_t pointers directly. There is no way to use keyboard and mousewheel before.

kisvegabor:
What if we add the indev to the default group only  if there is an already set default group?

I'm asking this because it's not sure that everybody wants to use the keyboard and encoder by default.

MouriNaruto:
@kisvegabor 
> What if we add the indev to the default group only if there is an already set default group?

I have export the input device objects (`lv_win32_pointer_device_object`, `lv_win32_keypad_device_object` and `lv_win32_encoder_device_object`) public and add a util function (`lv_win32_add_all_devices_to_group`) to make this implementation more customizable. I also do some modifications for tidying the implementation.

Note: The way to add all input devices which win32drv supported to the default group.
```
    if (!lv_win32_init(
        GetModuleHandleW(NULL),
        SW_SHOW,
        800,
        480,
        LoadIconW(GetModuleHandleW(NULL), MAKEINTRESOURCE(IDI_LVGL))))
    {
        return -1;
    }

    lv_group_t* default_group = lv_group_create();
    if (default_group)
    {
        lv_win32_add_all_devices_to_group(default_group);
        lv_group_set_default(default_group);
    }
```

win32drv/win32drv.c
```diff
@@ -182,6 +186,17 @@ static WPARAM volatile g_keyboard_value = 0;
  *   GLOBAL FUNCTIONS
  **********************/
 
+EXTERN_C void lv_win32_add_all_input_devices_to_group(
+    lv_group_t* group)
+{
+    if (group)
```
kisvegabor, line 192: Just nitpicking: It seems like a possible use case to not assign any groups (set `NULL`)  on all input devices.  What do you think?

Regardless of this, I like this new approach.

MouriNaruto: I have do some upgrade and hope it will be OK.

PR #139, opened 2021-07-29
Add some improvements for the new native Windows driver (win32drv)

MouriNaruto:
- Improve the zoomed graphic quality.
- Improve the DPI support.
- Fix out of range issue in the pointer input device driver.
- Add native render support for 16, 8 and 1 LVGL color depth mode.

Before

image

After

image

### DIFF
--- a/win32drv/win32drv.c
+++ b/win32drv/win32drv.c
@@ -13,6 +13,7 @@
 
 #include <windowsx.h>
 #include <VersionHelpers.h>
+#include <ShellScalingApi.h>
 
 #include <stdbool.h>
 #include <stdint.h>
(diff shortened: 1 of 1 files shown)


kisvegabor:
As far as I can judge it looks good. 

Side note: I didn't know the `MulDiv` functions. What a great name for this common calculation. Probably I'll add something like this in `lv_math.h`. :slightly_smiling_face:

MouriNaruto:
@kisvegabor 

Here is the documentation of 'MulDiv' function.
https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-muldiv

The function multiplies two 32-bit values and then divides the 64-bit result by a third 32-bit value. The final result is rounded to the nearest integer. It's a common API used in the High DPI support.

kisvegabor:
Thanks! In LVGL something like this is used to avoid floating-point numbers. E.g. `(10 * 3) / 4`.

MouriNaruto:
@kisvegabor 

> In LVGL something like this is used to avoid floating-point numbers.

But I do some search and don't find the similar API in lv_math.h.

kisvegabor:
>> Probably I'll add something like this in lv_math.h. slightly_smiling_face
 
> But I do some search and don't find the similar API in lv_math.h.


I meant I'll add it in the future. 

Anyway, as the PR looks good, so I merge it.

MouriNaruto:
Thanks for merging.

> I meant I'll add it in the future.

Got it.

PR #149, opened 2021-08-31
using lv_gpu_sdl

mariotaku:


display/monitor.c
```diff
@@ -347,42 +202,15 @@ static void window_create(monitor_t * m)
...
-    SDL_SetTextureBlendMode(m->texture, SDL_BLENDMODE_BLEND);
-
-    /*Initialize the frame buffer to gray (77 is an empirical value) */
-#if MONITOR_DOUBLE_BUFFERED
-    SDL_UpdateTexture(m->texture, NULL, m->tft_fb_act, MONITOR_HOR_RES * sizeof(uint32_t));
-#else
-    m->tft_fb = (uint32_t *)malloc(sizeof(uint32_t) * MONITOR_HOR_RES * MONITOR_VER_RES);
-    memset(m->tft_fb, 0x44, MONITOR_HOR_RES * MONITOR_VER_RES * sizeof(uint32_t));
-#endif
+    m->disp = lv_sdl_display_init(m->window);
```
kisvegabor, line 205: `lv_sdl_display_init` should be part of the driver not `lvgl`.

mariotaku: Do you think actually it's a good idea to put everything under `gpu/sdl` to `lv_drivers`, or a separate repo?

kisvegabor: I imagined it like this:
- monitor.c in  lv_dirvers:
  - `monitor_init()` initialize SDL, create/init "frame buffer", etc 
  -  `monitor_flush_cb()` just a flush callback that is assigned by the user
- gpu/sdl
  - all the drawing related stuff
- main.c
  - user creates a display driver and uses `monitor_flush_cb`
  
What do you think?

mariotaku: In my implementation, it's actually easier IMO. User creates SDL_Window with their preferred size, and use `lv_sdl_display_init` to initialize `Renderer` and `Texture` (act as frame buffer), and `lv_sdl_display_deinit` to clean them up. https://github.com/mariotaku/lvgl-sdl-example this is how I implement it.

kisvegabor: My problem with that is [these lines](https://github.com/mariotaku/lvgl-sdl-example/blob/64ba7d9fdef8e557116cd5fc560fa6b4d38b290a/main.c#L139-L149) requires some understanding of SDL, reading LVGL docs what needs to be initialized, how etc. For (not only) novice users we should to provide a "call an init and ready" complexity interface.

mariotaku: Good point. You mean if possible, you wish to expose less SDL related code to users as possible. Right?

kisvegabor: Yes, ideally no SDL API call. Do you think it's doable this GPU update too?

mariotaku: I think what's in lv_drivers is the best I can get. If a user wants to use lvgl with SDL, they should either use source from `monitor.c` to have a quick start, or learn SDL basics and add lvgl by themselves. As GPU driver only handles graphics, and for input and event handling one must write their own implementations to suit their need.

Another idea is to provide another example for SDL based driver. Where I can trim the code down to minimal.
By the way, I was thinking would it be better if we put all external renderer implementations into separate libraries.

kisvegabor: > they should either use source from monitor.c to have a quick start

If we can make an "init + set flush_cb" interface I'm happy. :slightly_smiling_face: 
So why can't we create the `SDL_Window` in `monitor.c` using `MONITOR_HOR/VER_RES` defines from `lv_drv_conf.h`?

mariotaku: > an "init + set flush_cb" interface
I'm not very sure about this one, now callbacks are implemented inside lv_gpu_sdl.c and users are not required to touch anything inside `lv_disp_drv_t`.

> create the SDL_Window in monitor.c using MONITOR_HOR/VER_RES defines from lv_drv_conf.h
I think we are doing this now according to monitor.c.

Maybe I missed something because I feel our desired design of public API is kinda different. 
Is the change in https://github.com/mariotaku/lv_sim_vscode_sdl/blob/master/main/src/main.c look good to you? If so, I'll be focusing on modifying monitor.c to fit structure of main source.

kisvegabor: My dream would be rather something like this: https://github.com/lvgl/lv_sim_eclipse_sdl/blob/master/main.c#L135-L143

Probably it's more complicated but mainly I mean moving the content of  `lv_sdl_display_init` to `monitor.c`.
Is there a dependency such as the the GPU driver in LVGL needs to see the same render, window, texture object that is initialized in the driver? If so I'd rather add a function to the SDL GPU in lvgl like `lv_gpu_sdl_set_texture()` which is called from `monitor_init`. What do you think?

mariotaku: I see. It doesn't sound impossible to me. However there are several caveats:

1. `SDL_Renderer` pointer is used everywhere, so far it has to be set by `user_data` field.
2. Any `SDL_Texture` are only usable by the `Renderer` who creates it.
3. A texture with screen size is created and set to `lv_disp_draw_buf_t` as `buf1`, which is a bit abusing the design by me.
4. In this `lv_sdl_display_init` there are 2 other essential calls: `lv_gpu_sdl_utils_init` and `lv_gpu_sdl_texture_cache_init` (and their deinit functions).

I myself is OK with doing all those above on my own, but I'm afraid it would increase learning curve if one wants to integrate SDL with HW acceleration.

kisvegabor: I see, thanks for describing these caveats. From this it seems it's far more complicated than only setting a `flush_cb`. 

What about creating a `monitor_gpu_init()` function that handles all the display creation stuff and leave the original `monitor_init()` for backward compatibility. 

BTW, as we already have various display driver for desktop, `monitor` is a bad name. It'd be better the move `monitor`, `keyboard`, `mouse`, `mousewheel` into `sdl.c/h` files. Similarly to [wayland](https://github.com/lvgl/lv_drivers/tree/master/wayland) for example. In this case we can leave `monitor_*` functions as wrappers to `sdl_init()` and `sdl_disp_flush()`. 

If you agree with this reorganization, I can do it and after that you can add a function like `sdl_gpu_create_display` (that handles everything).  

What do you think?

kisvegabor: > I'd be more than happy to get your help to reorganize the code. Yes, please proceed and let me know when I need to do my part.

Sorry for the late reply. I'll add this tomorrow (Tuesday).


> In the meantime I'll be doing some real world development using this branch to test memory and CPU usage, also bring some more proposals/patches :)

Amazing! :)

mariotaku: Several goodies mentioned above: 
* Implemented a recycled grid view, like GridView in Android
* Implemented something like Fragment in Android
* Image loading mechanism

I'll polish them, and contribute to lvgl repo if needed.

kisvegabor: I've added the driver reorganization. See [here](https://github.com/lvgl/lv_drivers/tree/master/sdl).

IMO it'd be better to add new `sdl_gpu.c/h` files for you changes. 

> Several goodies mentioned above:


Looks great, let me know if you have any LVGL related questions. :slightly_smiling_face:

mariotaku: @kisvegabor Do you wish me to move higher level drawing functions to this repo as well?

kisvegabor: IMO the the drawing functions should remain part of lvgl.

display/monitor.c
```diff
@@ -21,6 +21,8 @@
 #include "../indev/keyboard.h"
 #include "../indev/mousewheel.h"
 
+#include "../../lvgl/src/gpu/lv_gpu_sdl.h"
```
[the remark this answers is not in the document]
mariotaku: Usage to internal APIs has been all removed :)

mariotaku:
I have done changing the implementation. Dual monitor isn't supported yet, and there are 2 dirty hacks that's needed:

* Replacing `lv_disp_draw_buf_init` with `sdl_disp_draw_buf_init`
* Replacing `lv_disp_drv_init` with `sdl_disp_drv_init`

Please let me know if anything needs to be changed.

kisvegabor:
As I've suggested in [this comment](https://github.com/lvgl/lv_drivers/pull/149#discussion_r703563465) IMO it'd better to create a new `sdl_gpu.c/h`. There we can have an `lv_disp_t * sdl_gpu_init()` function, that just `malloc`s an `lv_disp_drv_t`, registers a display and  returns it. 
What do you think?

mariotaku:
@kisvegabor My bad, I missed that. I'll create a new driver for that.

mariotaku:
Done. Now it defaults to normal SDL implementation, and can be configured to use sdl_gpu.

kisvegabor:
Thanks! 
Now how can one initialize the driver? Only set the `flush_cb` and that's all?

sdl/sdl_gpu.h
```diff
@@ -0,0 +1,105 @@
...
+
+/**********************
+ * GLOBAL PROTOTYPES
+ **********************/
+
+/**
+ * Initialize SDL to be used as display, mouse and mouse wheel drivers.
+ */
+void sdl_init(void);
+
```
kisvegabor, line 48: Please a comment to this  `sdl_gpu_disp_drv_init` functions and emphasize that they need to be used instead of LVGL's `lv_disp_draw_buf_init` and `lv_disp_drv_init`.

kisvegabor:
Got it, then they should be comments to tell the user they should use drv_init and buf_ionit function from here.

sdl/sdl_gpu.h
mariotaku: Added a bit more description.

kisvegabor: Great thanks.

kisvegabor:
Looks good to me. IMO we should merge it together with the the lvgl part of SDL GPU update.

mariotaku:
For https://github.com/lvgl/lvgl/pull/2484 It's mostly stabilized.

kisvegabor:
lvgl/lvgl#2484 was merged so merging this too. Thank you very much again!

PR #150, opened 2021-08-31
Add support for pointer devices to libinput driver

Johennes:
This allows to use pointer devices such as mice or trackpads analogous to touch screens with the libinput driver.

I've found this helpful because it allowed me to develop embedded UIs on my desktop machine using my mouse or trackpad instead of having to do it on the touchscreen device directly.

Note that this touches the same file as #152 so one of the PRs will have to be rebased after the other is merged (if any of them is merged that is :crossed_fingers: :slightly_smiling_face:).

### DIFF
--- a/indev/libinput.c
+++ b/indev/libinput.c
@@ -169,6 +169,7 @@ report_most_recent_state:
  */
 static void read_pointer(struct libinput_event *event) {
   struct libinput_event_touch *touch_event = NULL;
+  struct libinput_event_pointer *pointer_event = NULL;
   enum libinput_event_type type = libinput_event_get_type(event);
   switch (type) {
     case LIBINPUT_EVENT_TOUCH_MOTION:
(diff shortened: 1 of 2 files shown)


kisvegabor:
Looks good, thank you!

Only one thing: in v8 the read_cb return `void` instead of `bool`.

Johennes:
> Only one thing: in v8 the read_cb returns `void` instead of `bool`.

Ah sorry, I missed that warning. Should be `void` now.

kisvegabor:
I've merged #152 so as you mentioned it needs some fixes.

Johennes:
Great! Thanks for merging it this way around because that meant less code to move around. :slightly_smiling_face: 

I rebased it on `master` again.

kisvegabor:
Looks good, thank you! :slightly_smiling_face:

PR #152, opened 2021-09-01
Add support for keypads to libinput driver

Johennes:
This allows to use keyboard / keypad devices with the libinput driver.

The implementation is largely analogous to the existing keypad support in the evdev driver. Unfortunately, I had to introduce a `USE_BSD_LIBINPUT` macro (again, analogous to the evdev driver) in order to get the right include path based on the system type.

Note that this touches the same file as #150 so one of the PRs will have to be rebased after the other is merged (if any of them is merged that is :crossed_fingers:  :slightly_smiling_face:).

### DIFF
--- a/indev/libinput.c
+++ b/indev/libinput.c
@@ -7,7 +7,7 @@
  *      INCLUDES
  *********************/
 #include "libinput_drv.h"
-#if USE_LIBINPUT != 0
+#if USE_LIBINPUT || USE_BSD_LIBINPUT
 
 #include <stdio.h>
 #include <unistd.h>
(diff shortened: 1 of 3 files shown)


kisvegabor:
Hi,

IMO `libinput_read` should be separated into 2 functions:
1. `libinput_read_touch`
2. `libinput_read_keyboard`

It's because `data->state` is the same for both type. Eg if `data->point={30,50}`, and you press the `g` key, `data->state` will be pressed and released, and LVGL will think the point {30,50} is also clicked.

If there are 2 read function the data and state will be separated. 

In the evdev driver it works because it sets `data` based on  `drv->type`.

Johennes:
> If there are 2 read function the data and state will be separated.

Good point! I have split up `libinput_read` to do the reading for touch and keypad devices in two separate functions now. This will need a little rebasing once #150 lands but basically just moving lines around.

kisvegabor:
Looks good, merging.

Thank you!

Issue #153, opened 2021-09-01
evdev / libinput: "full" keyboard support


Johennes:
When using the `LV_INDEV_TYPE_KEYPAD` type, the evdev driver supports a limited number of [keys](https://github.com/lvgl/lv_drivers/blob/master/indev/evdev.c#L174) (backspace, enter, up right, left, down). Any other keys are ignored. It would be great if the driver would handle input from a regular keyboard so that, for instance, typed characters are forwarded into a textarea.

I'm not sure if there is a better way to achieve this other than adding further cases for the relevant [key codes](https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h#L105) and handling things like SHIFT manually.

Maybe ASCII characters or characters found on common US English keyboard layouts would be enough for a start as that would limit the amount of `switch`ing.

I'm curious if you'd be open to adding this to lv_drivers and would be happy to help implement it.

kisvegabor:
For "normal letters" `data->key = in.code;` should work instead of `data->key = 0;` in the `default` branch of the `switch case`.

It really works only for ASCII letters. Isn't there a way to get the "real" letters, e.g. `shift + a = A`?

Johennes:
It seems that [libxkbcommon](https://xkbcommon.org/doc/current/modules.html) would be able to do the translation from key codes to key symbols based on the current key map. I'll have to do some more research to understand what would be required to hook into that.

Johennes:
Looks like libxkbcommon is working pretty ok for me so far. We'd only have to add a config setting for the key map to use. E.g. this is what I use with my laptop keyboard:

```
struct xkb_rule_names names = {
    .rules = NULL,
    .model = "pc105",
    .layout = "de",
    .variant = "nodeadkeys",
    .options = NULL
};
```

kisvegabor:
Good news! :slightly_smiling_face: 

> We'd only have to add a config setting for the key map to use.

Shall we add parts of this config to `lv_drv_conf.h`?

Johennes:
Yeah, good idea. :+1:

Let me play with it some more and then I'll try to put it into a pull request.

Issue #151, opened 2021-08-31
evdev / libinput: use multiple devices


Johennes:
The current evdev and libinput drivers store state in static variables. That means they can only be used with one device at a time but I cannot use, say, two trackpads or a trackpad and a keyboard at the same time (or at least my weak C-fu doesn't let me see how).

It would be great if the driver state could instead be stored in an external structure, to be managed by the caller. This way I could connect an arbitrary number of input devices via either of these drivers.

In order to not break the interface of the existing evdev / libinput drivers maybe a new driver with support for multiple devices could be created and the existing drivers could be changed to use the new one under the hood.

Johennes:
Actually we wouldn't even need the state to be managed externally. We could instead store it in the driver's [`user_data`](https://github.com/lvgl/lvgl/blob/master/src/hal/lv_hal_indev.h#L101) field. I made a hacky proof-of-concept [here](https://gitlab.com/cherrypicker/unl0kr/-/blob/master/libinput_multi.c).

kisvegabor:
I agree to support various type of indevs (mouse+touch+keypad) but I can't imagine a use case for e.g. 2 keyboard, or 2 mouses. 

If we target only 1 indev per type we can simply hardcode `state_mouse`, `state_touch`, etc.

What do you think?

Johennes:
Ah, that's an interesting thought. I suppose that would get me half of the way at least.

There are two cases I can think of where being able to connect multiple devices of the same type might be helpful:

1.A device with built-in keyboard (for instance, a Pinebook) that is also docked to an external USB keyboard
2. Any device docked to an external trackpad and a keyboard with mouse emulation. For example, I have an Ergodox EZ and I occasionally like to control the mouse with it but most of the time I use a trackpad. The keyboard's mouse emulation shows up as a separate device though.

```
/dev/input/event8:      Apple Inc. Magic Trackpad 2
/dev/input/event9:      ZSA Ergodox EZ
/dev/input/event10:     ZSA Ergodox EZ Mouse
/dev/input/event11:     ZSA Ergodox EZ System Control
/dev/input/event12:     ZSA Ergodox EZ Consumer Control
/dev/input/event13:     ZSA Ergodox EZ Keyboard
```

Both of these scenarios are probably not overly common though so if you think it'd add too much complexity to allow for multiple devices of the same type, just permitting one device per type might be enough. On the other hand, the step from that solution to allowing an arbitrary number of devices might also not be too big anymore. :thinking:

[...]

PR #156, opened 2021-09-04
Add full keyboard support to libinput/evdev driver

Johennes:
This enhances the existing libinput and evdev driver by adding support for full keyboards. The conversion from Linux scan codes to UTF-8 characters is performed via XKB. To control the behavior a new config setting for opting into or out of using XKB was added. The keymap can either be set via another config option or by calling a function at runtime.

Two small caveats:

1. This only works for a single connected keyboard because static variables are used to store the keyboard state. Supporting multiple devices of the same type is still being discussed in #151.
2. When entering non-ASCII characters, the selected font needs to include the respective glyph, otherwise it won't render on the UI.

Fixes: #153

### DIFF
--- a/indev/evdev.c
+++ b/indev/evdev.c
@@ -18,6 +18,10 @@
 #include <linux/input.h>
 #endif
 
+#if USE_XKB
+#include "xkb.h"
+#endif /* USE_XKB */
+
 /*********************
  *      DEFINES
  *********************/
@@ -74,6 +78,10 @@ void evdev_init(void)
     evdev_root_y = 0;
     evdev_key_val = 0;
     evdev_button = LV_INDEV_STATE_REL;
+
+#if USE_XKB
+    xkb_init();
+#endif
 }
 /**
  * reconfigure the device file for evdev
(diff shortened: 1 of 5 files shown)


indev/evdev.c
```diff
@@ -170,42 +178,55 @@ void evdev_read(lv_indev_drv_t * drv, lv_indev_data_t * data)
...
+                data->key = xkb_process_key(in.code, in.value != 0);
+#else
+                switch(in.code) {
+                    case KEY_BACKSPACE:
+                        data->key = LV_KEY_BACKSPACE;
+                        break;
+                    case KEY_ENTER:
+                        data->key = LV_KEY_ENTER;
+                        break;
+                    case KEY_PREVIOUS:
```
kisvegabor, line 191: Which keys are `KEY_PREVIOUS` and `KEY_NEXT` are mapped to `-` and `+`, right?

For `LV_KEY_NEXT` `tab` would be more conventional to move the focus.

kisvegabor (commented):
Thank you!

I haven't tested it yet, but look good in general. Only a minor comment.

indev/evdev.c
Johennes: Hm, I'm actually not sure what keys trigger `KEY_PREVIOUS` and `KEY_NEXT`. At least none of the keys on my keyboards seem to. I just found them in the [Linux event code header](https://github.com/torvalds/linux/blob/master/include/uapi/linux/input-event-codes.h#L479) and thought they'd be good to map to `LV_KEY_PREV` and `LV_KEY_NEXT`, too. Googling around I found a [forum thread](https://www.vdr-portal.de/forum/index.php?thread/126665-tbs6982-codes-umbiegen-key-previous-key-back/) where somebody receives these keys from an IR remote.

Maybe we could map all of `KEY_RIGHT` / `KEY_NEXT` / `KEY_TAB` to `LV_KEY_NEXT` / `KEY_LEFT` and `KEY_PREVIOUS` to `LV_KEY_PREV`?

kisvegabor: Not that, `LV_KEY_NEXT` and `LV_KEY_LEFT` are different.
- `LV_KEY_NEXT`: focus on the next object
- `LV_KEY_LEFT`: like left arrow, move the cursor on a text area, adjust slider, etc. 


So I suggest binding 
- `KEY_LEFT/RIGHT/UP/DOWN` to `LV_KEY_LEFT/RIGHT/UP/DOWN` 
- `KEY_TAB` to `LV_KEY_NEXT`
- Nothing to `LV_KEY_PREV` (Shift+Tab could be good but I don't if how difficult it is to handle)

Johennes: Ah, got it. Ok, I pushed a change to:

- map `KEY_TAB` to `LV_KEY_NEXT` _outside_ the XKB path
- map XKB's tab and "left tab" (sent on Shift + Tab) key symbols to `LV_KEY_NEXT` and `LV_KEY_PREV`, respectively

I did not add handling for Shift + Tab _outside_ the XKB path because there's currently no handling of Shift at all there. It wouldn't be hard but if somebody needs Shift + Tab (or capital letters), they're probably better off with `#define USE_XKB 1` anyway.

While at it, I noticed that XKB has separate key symbols for keys pressed on the key/number pad. E.g. there is `XKB_KEY_Return` and also `XKB_KEY_KP_Enter`. I simply added these to the existing cases.

I also left the mapping of `XKB_KEY_Prior` and `XKB_KEY_Next` to `LV_KEY_PREV` and `LV_KEY_NEXT` in. If somebody really has an input device that sends these keys, I suppose that would be the most reasonable mapping.

Let me know if this is ok and I'm happy to adapt it if needed.

kisvegabor:
All looks good now, merging, thank you!

Issue #155, opened 2021-09-03
libinput: automatic device discovery


Johennes:
I'm working on an application that is meant to be run on many different types of devices which means I can't hard-code the `/dev/input/event...` paths because I don't know them at build time. It would be great if the libinput driver could provide an interface for automatically detecting touchscreens, pointer devices and keyboards.

I have some very rudimentary code [here](https://gitlab.com/cherrypicker/unl0kr/-/blob/master/libinput_device_discovery.c#L66) that basically just walks the `/dev/input/event...` files, makes a libinput call to determine the device capabilities and then filters the files based on that.

I'd be interested to know if you would consider adding something like this to lv_drivers? I think this could be beneficial to other users as well.

kisvegabor:
What if you simply do `#define LIBINPUT_NAME   my_find_input_dev()`? 

Anyway, I believe it'd be useful to add `libinput_find_touch_dev()`, `libinput_find_mose_dev()`, etc  to `libinput.c` so that the user could use them as `LIBINPUT_NAME`. 

Probably we'd need `LIBINPUT_TOUCH_NAME`, `LIBINPUT_MOUSE_NAME`, etc.

What do you think?

Johennes:
Yeah, this overlaps a little with #151.

Using it directly in the config file is an interesting idea! The functions I'm using so far return all devices of a certain type. We could maybe wrap them in another set of functions to return only the first  device of a type, that shouldn't be too complicated.

kisvegabor:
> Using it directly in the config file is an interesting idea! The functions I'm using so far return all devices of a certain type. We could maybe wrap them in another set of functions to return only the first device of a type, that shouldn't be too complicated.

Based on my [last comment](https://github.com/lvgl/lv_drivers/issues/151#issuecomment-913202619) in #151, I think it'd be the approach now.

PR #157, opened 2021-09-05
Automatic device discovery via libinput

Johennes:
This adds new convenience functions to the libinput driver for automatically detecting input devices of a certain type (such as keyboards or touchscreens).

My pointer magic is weak so I hope I did everything correctly here. Looking forward to candid feedback. :slightly_smiling_face: 

The logic could also be generalized to work with the evdev driver, too. That would just require replacing the libinput calls with `ioctl`s to the kernel. Since I'm not currently using the evdev driver, I didn't take the time to do that but it should be relatively straightforward for somebody else to do if they ever need it.

Closes: #155

### DIFF
--- a/indev/libinput_drv.h
+++ b/indev/libinput_drv.h
@@ -36,11 +36,35 @@ extern "C" {
 /**********************
  *      TYPEDEFS
  **********************/
+typedef enum {
+  LIBINPUT_CAPABILITY_NONE     = 0,
+  LIBINPUT_CAPABILITY_KEYBOARD = 1U << 0,
+  LIBINPUT_CAPABILITY_POINTER  = 1U << 1,
+  LIBINPUT_CAPABILITY_TOUCH    = 1U << 2
+} libinput_capability;
 
 /**********************
  * GLOBAL PROTOTYPES
  **********************/
 
+/**
+ * find connected input device with specific capabilities
+ * @param capabilities required device capabilities
+ * @param force_rescan erase the device cache (if any) and rescan the file system for available devices
+ * @return device node path (e.g. /dev/input/event0) for the first matching device or NULL if no device was found.
+ *         The pointer is safe to use until the next forceful device search.
+ */
+char *libinput_find_dev(libinput_capability capabilities, bool force_rescan);
+/**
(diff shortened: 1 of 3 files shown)


indev/libinput_drv.h
```diff
@@ -36,11 +36,35 @@ extern "C" {
 /**********************
  *      TYPEDEFS
  **********************/
+enum libinput_capability {
```
kisvegabor, line 39: Usually we use `typedef enum`s. Can we use it here too?

kisvegabor (commented):
Looks good, only one minor comment.

indev/libinput_drv.h
Johennes: Ah, yes absolutely. That makes all the call sites a little more brief. :+1: 

Made the change.

kisvegabor: Thank you! :slightly_smiling_face:

lv_drv_conf_template.h
```diff
@@ -384,6 +384,9 @@
 
 #if USE_LIBINPUT || USE_BSD_LIBINPUT
 #  define LIBINPUT_NAME   "/dev/input/event0"        /*You can use the "evtest" Linux tool to get the list of devices and test them*/
+
+/*If only a single device of the same type is connected, you can also auto detect it, e.g.:
+ *#  define LIBINPUT_NAME   libinput_find_dev(LIBINPUT_CAPABILITY_TOUCH, false)*/
```
kisvegabor, line 389: Please move this comment above `#  define LIBINPUT_NAME   "/dev/input/event0"`.

kisvegabor (commented):
I've just found an other minor thing. 

In light of the other PRs can we merge it?

lv_drv_conf_template.h
Johennes: Good idea. Made the change.

Johennes:
> In light of the other PRs can we merge it?

#156 probably needs minor rebasing once this one has landed but shouldn't be too bad, I think.

kisvegabor:
Looks good, thank you!

PR #158, opened 2021-09-05
fix(monitor) Bug setting w/h in SDL renderer rect

kevinpt:
This only affects builds using LV_COLOR_SCREEN_TRANSP


kisvegabor:
Thanks you!

PR #159, opened 2021-09-07
added MONITOR_FULLSCREEN option

AIWintermuteAI:
While developing demo for Seeed Studio reTerminal
https://github.com/AIWintermuteAI/Seeed_reTerminal_LVGL_UI_Demo
I needed to enable full screen mode for SDL, instead of windowed.
Hence the option (with ifndef guard for preserving compatibility with older configs) was added.


kisvegabor:
Hi,

There  was minor refactoring meanwhile and `monitor` was moved [here](https://github.com/lvgl/lv_drivers/blob/master/sdl/sdl.c). Could you resend the PR?

display/monitor.c
```diff
@@ -345,7 +349,7 @@ static void window_create(monitor_t * m)
 {
     m->window = SDL_CreateWindow("TFT Simulator",
                               SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED,
-                              MONITOR_HOR_RES * MONITOR_ZOOM, MONITOR_VER_RES * MONITOR_ZOOM, 0);       /*last param. SDL_WINDOW_BORDERLESS to hide borders*/
+                              MONITOR_HOR_RES * MONITOR_ZOOM, MONITOR_VER_RES * MONITOR_ZOOM, MONITOR_FULLSCREEN);       /*last param. SDL_WINDOW_BORDERLESS to hide borders*/
```
kisvegabor, line 352: It seems accidentally the SDL flag fro full screen is 1. It'd better to use the `SDL_WINDOW_FULLSCREEN` flag is required.

AIWintermuteAI: Hmmm... It unlikely to change, but in case it does it is safer indeed to use the flag. I came up with this 
```
   #if SDL_FULLSCREEN
    int full_scr_flag = SDL_WINDOW_FULLSCREEN;
    #else
    int full_scr_flag = 0;
    #endif

    m->window = SDL_CreateWindow("TFT Simulator",
                              SDL_WINDOWPOS_UNDEFINED, SDL_WINDOWPOS_UNDEFINED,
                              SDL_HOR_RES * SDL_ZOOM, SDL_VER_RES * SDL_ZOOM, full_scr_flag);       /*last param. SDL_WINDOW_BORDERLESS to hide borders*/
```
looks a bit clumsy though. Perhaps there is a better way or this is fine?

kisvegabor: I can't imagine a better significantly better way. A  minor thing but I'd do:
```c
int flag = 0
#if SDL_FULLSCREEN
flag |= SDL_WINDOW_FULLSCREEN;
#endif
```

This way we can easily add other flags later.

AIWintermuteAI:
Okay, added the discussed changes!

kisvegabor:
Looks good, thank you!

PR #160, opened 2021-09-07
fix(evdev) document parameter evdev_read returns void

ckielstra:
evdev_read was changed to return void, but the parameter descriptions were not updated accordingly.


kisvegabor:
Thank you!

Issue #161, opened 2021-09-07
evdev / libinput: support KEY_POWER


Johennes:
Some devices have dedicated power buttons that emit the `KEY_POWER` scan code. Here's an example from the power button on my laptop:

```
Input driver version is 1.0.1
Input device ID: bus 0x19 vendor 0x0 product 0x1 version 0x0
Input device name: "Power Button"
Supported events:
  Event type 0 (EV_SYN)
  Event type 1 (EV_KEY)
    Event code 116 (KEY_POWER)
Properties:
Testing ... (interrupt to exit)
Event: time 1631010379.405744, type 1 (EV_KEY), code 116 (KEY_POWER), value 1
Event: time 1631010379.405744, -------------- SYN_REPORT ------------
Event: time 1631010379.405837, type 1 (EV_KEY), code 116 (KEY_POWER), value 0
Event: time 1631010379.405837, -------------- SYN_REPORT ------------
```

As an addendum to #153 and #155, it would be nice if:

1. the libinput driver could automatically recognize power-button devices
2. the evdev / libinput drivers could handle `KEY_POWER`

For 1, we'd just have to add another case to the `libinput_capability` enum and a corresponding capability check (just like we do for the other device types in #157).

For 2, performing the shutdown is as easy as

```
#include <sys/reboot.h>
sync();
reboot(RB_POWER_OFF);
```

However, I am not sure how to best integrate this into the existing drivers. :thinking: 

We could add a new case to `LV_INDEV_TYPE_POWER` to the [`lv_indev_type_t`](https://github.com/lvgl/lvgl/blob/master/src/hal/lv_hal_indev.h#L68) enum in lvgl but that might mean quite invasive changes.

Alternatively, we could just use `LV_INDEV_TYPE_KEYPAD` for power button devices and handle the key in [`indev_keypad_proc`](https://github.com/lvgl/lvgl/blob/f61b2ca45502472cde8ac0983b73dbf153de2b20/src/core/lv_indev.c#L411). It'll feel a little odd to refer to the device as "keypad" but after all the kernel treats it as a keyboard, too, because it can send keys so maybe that's the best approach.

kisvegabor:
Interesting idea but I like it. It sounds like a generic method to shout down a device.

[...]

PR #163, opened 2021-09-16
Add support for devices with KEY_POWER

Johennes:
This makes it possible to automatically locate power buttons (which have `KEY_POWER)` via libinput. Additionally, a new `LV_KEY_POWER` key is introduced that is dispatched whenever KEY_POWER is pressed.

In order to handle `LV_KEY_POWER,` the client needs to register an indev  with type `LV_INDEV_TYPE_KEYPAD` and set the `feedback_cb` callback. Additionally, the indev needs to be assigned a group with at least one element.

**Note:** This is a draft and I'd like to get feedback on it. The fact that a group has to be assigned feels a little odd but that was the only way to make `feedback_cb` be called. After having tried it out, I'm not sure if this is a good way to implement support for `KEY_POWER`. Maybe it'd be better if the client would supply a custom `read_cb` and handle the `KEY_...` events manually because in the case of `KEY_POWER`, we don't really need any of the group logic?

Fixes: #161

### DIFF
--- a/indev/evdev.c
+++ b/indev/evdev.c
@@ -208,6 +208,9 @@ void evdev_read(lv_indev_drv_t * drv, lv_indev_data_t * data)
                     case KEY_TAB:
                         data->key = LV_KEY_NEXT;
                         break;
+                    case KEY_POWER:
+                        data->key = LV_KEY_POWER;
+                        break;
                     default:
                         data->key = 0;
                         break;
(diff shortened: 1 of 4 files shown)


kisvegabor:
>  Additionally, the indev needs to be assigned a group with at least one element.

Uh, it's really odd. I forgot about it.  Usually, if there is a `KEYPAD` indev there are group(s) too but it's really no that logical. 


  > Maybe it'd be better if the client would supply a custom read_cb and handle the KEY_... events manually because in the case of KEY_POWER, we don't really need any of the group logic?

If we can't give clean solution IMO it's better to leave it to the user as you suggested.

Johennes:
> If we can't give clean solution IMO it's better to leave it to the user as you suggested.

Ok, that sounds fair. I'll close this.

PR #164, opened 2021-09-16
Make it possible to set the variable offset for fbdev

Johennes:
This adds a new function for setting the x and y offset for the framebuffer display driver. This is useful when the driver should only
occupy part of the screen, for instance for an on-screen keyboard, and leave the remainder of the screen untouched.


kisvegabor:
Seems useful, thank you!

PR #167, opened 2021-09-24
Use LV_LOG instead of printf in fbdev driver

Johennes:
This replaces the previous `printf` calls with `LV_LOG_INFO`. When using LVGL in an application that has to forward STDOUT to another process, this is required so that logging output from the fbdev driver can be controlled with the LV_LOG settings to prevent feeding it to other programs.

See also: https://gitlab.com/cherrypicker/unl0kr/-/issues/16

### DIFF
--- a/display/fbdev.c
+++ b/display/fbdev.c
@@ -92,7 +92,7 @@ void fbdev_init(void)
         perror("Error: cannot open framebuffer device");
         return;
     }
-    printf("The framebuffer device was opened successfully.\n");
+    LV_LOG_INFO("The framebuffer device was opened successfully");
 
     // Make sure that the display is on.
     if (ioctl(fbfd, FBIOBLANK, FB_BLANK_UNBLANK) != 0) {
@@ -138,7 +138,7 @@ void fbdev_init(void)
     }
 #endif /* USE_BSD_FBDEV */
 
-    printf("%dx%d, %dbpp\n", vinfo.xres, vinfo.yres, vinfo.bits_per_pixel);
+    LV_LOG_INFO("%dx%d, %dbpp", vinfo.xres, vinfo.yres, vinfo.bits_per_pixel);
 
     // Figure out the size of the screen in bytes
     screensize =  finfo.smem_len; //finfo.line_length * vinfo.yres;    
@@ -151,7 +151,7 @@ void fbdev_init(void)
     }
     memset(fbp, 0, screensize);
 
-    printf("The framebuffer device was mapped to memory successfully.\n");
+    LV_LOG_INFO("The framebuffer device was mapped to memory successfully");
 
 }
 


embeddedt (approved):

Johennes:
Sniper approve! :sweat_smile:

embeddedt:
Thanks!

> It looks like LVGL can currently only set its own log level at compile time so unless we want to upstream an enhancement for this, we'll have to turn its logging off entirely.

One solution is to allow `LV_LOG_LEVEL` to not be a constant, however, this would require some type of preprocessor incantation to detect it and skip [these checks](https://github.com/lvgl/lvgl/blob/28e9593e5802a2e7d493515059c6327e60ccbf28/src/misc/lv_log.h#L86-L114).

Johennes:
> One solution is to allow `LV_LOG_LEVEL` to not be a constant, however, this would require some type of preprocessor incantation to detect it and skip [these checks](https://github.com/lvgl/lvgl/blob/28e9593e5802a2e7d493515059c6327e60ccbf28/src/misc/lv_log.h#L86-L114).

Oh interesting. Would it be thinkable to push the `LV_LOG_LEVEL` comparison _into_ the macros?

So, for instance, instead of 

```
#if LV_LOG_LEVEL <= LV_LOG_LEVEL_TRACE
#define LV_LOG_TRACE(...) _lv_log_add(LV_LOG_LEVEL_TRACE, __FILE__, __LINE__, __func__, __VA_ARGS__)
#else
#define LV_LOG_TRACE(...) do {}while(0)
#endif
```

have

```
#define LV_LOG_TRACE(...) do { _lv_log_add(LV_LOG_LEVEL_TRACE, __FILE__, __LINE__, __func__, __VA_ARGS__); break; } while (LV_LOG_LEVEL <= LV_LOG_LEVEL_TRACE)
```

or something similar?

It looks like there are also two other places where the log level is checked in `#if` ([`lv_img.c`](https://github.com/lvgl/lvgl/blob/28e9593e5802a2e7d493515059c6327e60ccbf28/src/widgets/lv_img.c#L78) and [`lv_obj.c`](https://github.com/lvgl/lvgl/blob/28e9593e5802a2e7d493515059c6327e60ccbf28/src/core/lv_obj.c#L176)) that would probably need similar handling.

embeddedt:
The main issue with that solution is that I doubt all compilers will be
able to optimize the loop out,  even if the condition is known to always be
false.

Johennes:
Probably not, true. Unfortunately, my macro-fu is too weak to see a way of how we could make `LV_LOG_LEVEL` a variable _and_ keep a no-op when the value is higher than a specific level. It seems to me that if it's not a compile-time constant, we'd always have at least an `if` check. :disappointed:

PR #172, opened 2021-10-03
fix(sdl) update monitor texture pointer during resize

ReimuNotMoe:
Texture pointer in monitor struct should be updated during SDL window resize, otherwise the old texture pointer will still be in use and result in invalid memory accesses.

### DIFF
--- a/sdl/sdl_gpu.c
+++ b/sdl/sdl_gpu.c
@@ -214,6 +214,7 @@ void sdl_display_resize(lv_disp_t *disp, int width, int height)
     }
     SDL_Texture *texture = SDL_CreateTexture(renderer, SDL_PIXELFORMAT_ARGB8888, SDL_TEXTUREACCESS_TARGET, width,
                                              height);
+    monitor.texture = texture;
     lv_disp_draw_buf_init(driver->draw_buf, texture, NULL, width * height);
     driver->hor_res = (lv_coord_t) width;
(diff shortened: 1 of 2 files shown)


embeddedt:
This seems like a good fix to me. @mariotaku do you also agree?

mariotaku:
Indeed. Not sure if we should do anything to old texture pointer though.

ReimuNotMoe:
I want to add some more improvements to enable the end user to edit window title, handle more SDL events, etc. Is including SDL header allowed in `sdl_gpu.h`?

mariotaku:
IMO for advanced configuration, we can ask users to create their own window and renderer, and configure the disp driver by themselves. That how I did in my project: https://github.com/mariotaku/moonlight-tv/tree/main/app/lvgl

Handle display resize: https://github.com/mariotaku/moonlight-tv/blob/main/app/lvgl/lv_disp_drv_app.c
IME support: https://github.com/mariotaku/moonlight-tv/blob/main/app/lvgl/lv_sdl_drv_key_input.c (CJK input is perfectly fine)

ReimuNotMoe:
Thanks you the suggestions. I won't make more changes. You can merge the existing two commits if you want.

mariotaku:
LGTM!

embeddedt:
Thank you; merged.

PR #175, opened 2021-10-06
build: Remove the optional install flag

rzr:
We suspected it broke something on ESP32 but it was something else
So let's remove uneeded complexity

Relate-to: https://github.com/lvgl/lvgl/pull/2636
Signed-off-by: Philippe Coval <philippe.coval@huawei.com>


rzr:
May some build bots added here too

Cc: @embeddedt

PR #186, opened 2021-10-31
fix(CMake) add includes and dependencies

higaski:
- Changed include_directories (bad practice) to target_include_directories
- Added dependencies
  - lv_drivers depends on lvgl
- Namespaced target aliases
  - lvgl::drivers
- Applied cmake-format for a consistent style

This PR is also part of [#2753](https://github.com/lvgl/lvgl/pull/2753)


kisvegabor:
Thanks!

PR #189, opened 2021-11-08
normalize including lv_drv_conf.h

smarq8:
all files use "../../lv_drv_conf.h" but win_drv.h used "../lv_drv_conf.h"
screenshot 1636392173


smarq8:
I saw that but now I can see...

PR #190, opened 2021-11-09
fix(sdl) Default macro defs for SDL and XKB drivers

kevinpt:
Code referencing SDL_FULLSCREEN and USE_XKB needs these macros defined unconditionally
as it doesn't check their presence with `#if defined`.


kisvegabor:
Looks good, thanks!

PR #193, opened 2021-11-22
Refactor(evdev) removed duplicate code from init function

ckielstra:
Two almost identical functions were combined.

Small side effect is a bug fix for a perhaps never occurring use case; when evdev_set_file was being called without evdev_init, then this was closing a random device.


kisvegabor:
Looks good, merging.

Thanks!

PR #196, opened 2021-12-05
Wayland: add fullscreen support

WallaceIT:
Add support for setting windows as fullscreen with a new API call (`lv_wayland_window_set_fullscreen`).


kisvegabor:
Thanks you!

PR #200, opened 2022-03-17
fix(evdev): small warning about dangling else

markand:
warning: suggest explicit braces to avoid ambiguous 'else' [-Wdangling-else]
  151 |             else if(in.code == ABS_MT_TRACKING_ID)
      |


kisvegabor:
Looks reasonable, thank you!

PR #203, opened 2022-04-10
fix(wayland): Fix XDG surface buffer attach

simplejack-src:
When using XDG Shell, XDG surface buffer cannot be attached to XDG surface until after surface has been configured.


kisvegabor:
Thanks!
I couldn't test it but I assume tried it out and it worked :slightly_smiling_face:

PR #205, opened 2022-04-16
feat(cmake-build): Option to allow building shared libraries.

simplejack-src:
Add option in CMake build to specify that shared libraries should be created (as
opposed to static). This is an exclusive option (not enabled by default).

I'm by no means a CMake specialist (or even fan), but I believe this is the (current) recommend method.


kisvegabor:
Thanks you!

PR #212, opened 2022-05-01
Use direct mode instead of rounder callback in win32drv.

MouriNaruto:
For @FASTSHIFT's suggestion, win32drv started to use direct mode instead of rounder callback.

Also, win32drv can reduce one memory copy for each frame when Windows supports the color depth your set in LVGL.

P.S. When I use the profiler, Windows GDI operation uses 80% of the main thread CPU time. I will try to create a new PR to offload them to a separate thread.

Kenji Mouri


MouriNaruto:
@kisvegabor ping for progress

Kenji Mouri

PR #192, opened 2021-11-21
fix(libinput): read display dimensions from driver

Johennes:
This changes the libinput driver to read unrotated display dimensions directly from the driver. This is necessary because libinput itself doesn't respect any rotation inside of LVGL. As a result, coordinates retrieved from libinput need to be compared to the unrotated dimensions.

Fixes: #188

PR #214, opened 2022-05-09
fix(libinput) handle cases where physical resolutions are 0

g7:
When physical_hor_res and physical_ver_res aren't specified, use hor_res and ver_res as before #192.


embeddedt:
Thanks! The fix looks good.

PR #216, opened 2022-05-17
Offload window management and Windows GDI operation to a separate thread in win32drv.

MouriNaruto:
In https://github.com/lvgl/lv_drivers/pull/212, I said that I will try to create a new PR to offload them to a separate thread because Windows GDI operation uses 80% of the main thread CPU time when I use the profiler. Here is the code.

Before
image

After
image

Kenji Mouri

### DIFF
--- a/win32drv/win32drv.c
+++ b/win32drv/win32drv.c
@@ -12,7 +12,7 @@
 #if USE_WIN32DRV
 
 #include <windowsx.h>
-
+#include <process.h>
 #include <stdbool.h>
 #include <stdint.h>
 
@@ -37,6 +37,16 @@
 /**********************
  *      TYPEDEFS
  **********************/
+
+typedef struct _WINDOW_THREAD_PARAMETER
+{
+    HANDLE window_mutex;
+    HINSTANCE instance_handle;
+    HICON icon_handle;
+    lv_coord_t hor_res;
+    lv_coord_t ver_res;
+    int show_window_mode;
+} WINDOW_THREAD_PARAMETER, * PWINDOW_THREAD_PARAMETER;
 
 /**********************
  *  STATIC PROTOTYPES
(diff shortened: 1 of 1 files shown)


MouriNaruto:
@kisvegabor ping for progress

Kenji Mouri

kisvegabor:
Wow, it's unbelievable. Again :slightly_smiling_face:  

I started to work on the new driver architecture which will be part of the main repo. It consists of `drv`s (low level) and `dev`s (high level). In case of SDL, [SDL is a driver](https://github.com/lvgl/lvgl/blob/arch/drivers/src/drv/sdl/lv_drv_sdl_window.c) and [frame buffer](https://github.com/lvgl/lvgl/blob/arch/drivers/src/dev/disp/fb/lv_dev_fb.c) is an abstract device which uses the SDL driver. 

It can be used like this:
```c
  lv_drv_sdl_window_t * drv_win = lv_mem_alloc(sizeof(lv_drv_sdl_window_t));
  lv_drv_sdl_window_init(drv_win);
  drv_win->hor_res = w;
  drv_win->ver_res = h;
  lv_disp_drv_t * disp_drv = lv_drv_sdl_window_create(drv_win);

  lv_dev_fb_t * dev_fb = lv_mem_alloc(sizeof(lv_dev_fb_t));
  lv_dev_fb_init(dev_fb);
  dev_fb->hor_res = w;
  dev_fb->ver_res = h;
  lv_disp_t * disp = lv_dev_fb_create(dev_fb, disp_drv);
```

As I imagine Linux `/dev/fb`, your Windows driver, STM32's frame buffer, etc will be `drv`s hidden behind a a frame buffer `dev`.
Do you think the Windows driver can with into this scheme?

MouriNaruto:
@kisvegabor 

> Do you think the Windows driver can with into this scheme?

I am reading the source code you committed in the `arch/drivers` branch.

It looks like I can try to integrate the Windows driver to this scheme, but I need to try first and validate my guess.

Kenji Mouri

kisvegabor:
If it looks good at first look, it's already something. :slightly_smiling_face: 
I'm still also validating the architecture by adding some very different drivers. 

I'll let you know when it's stabilized.

kisvegabor:
For this callbacks I was considering an SPI based display controller (e.g. ILI9341) and an I2C touchpad controller. For these purpose of these functions is quite different:
- reset: set/clear GPIO
- send: send a `cmd` specified by the controller IC via SPI or I2C
- read: read some data via SPI or I2C
- send_img: the core of the flush_cb

Anyway, I keep this option in mind.

PR #217, opened 2022-05-27
Make it possible to query fbdev dpi

Johennes:
This computes the display's DPI from the variable fbdev info which makes this method analogous to `drm_get_sizes`.


kisvegabor:
Thank you!

PR #218, opened 2022-05-27
Fix warnings from Meson's warning level 2

Johennes:
Fixes three small warnings reported by activating warning level 2 in meson.


kisvegabor:
Thanks!

PR #220, opened 2022-06-14
sdl_common.c: Map SHIFT-TAB to LV_KEY_PREV

amirgon:


kisvegabor:
I wanted to add this for so long but never took the time to find how to get the modifier key :smile: 

Thank you!

PR #223, opened 2022-06-24
fix(wayland): Add a missing XDG shell event handler.

simplejack-src:
A new Wayland XDG shell protocol event was added upstream, so add a corresponding (but currently unused) handler.


kisvegabor:
Thank you!

PR #236, opened 2022-09-22
Make windows.h lower case since it is the correct filename

Viatorus:
Windows itself is case insensitve, but the correct filename of _windows.h_ is written in lower case. 

If the lv_drivers is built cross-compiled for Windows from Linux via MinGW, this leads to an error, since Linux is case sensitive.

See:
https://stackoverflow.com/questions/15466613/lowercase-windows-h-and-uppercase-windows-h-difference

### DIFF
--- a/win32drv/win32drv.h
+++ b/win32drv/win32drv.h
@@ -20,7 +20,7 @@
 
 #if USE_WIN32DRV
 
-#include <Windows.h>
+#include <windows.h>
 
 #if _MSC_VER >= 1200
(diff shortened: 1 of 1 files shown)


kisvegabor:
Thank you!

MouriNaruto:
@Viatorus 

In Windows SDK, the file name is `Windows.h`. So, `Windows.h` is not the wrong file name. (It's the MinGW's issue.)

But the change is good for doing workaround for MinGW.

Kenji Mouri

PR #237, opened 2022-10-14
fix(drm): Fix compiler warnings

Johennes:


display/drm.c
```diff
@@ -464,7 +470,7 @@ static int drm_find_connector(void)
 		drmModeFreeEncoder(enc);
 	}
 
-	drm_dev.crtc_idx = -1;
```
Johennes, line 467: -1 isn't correct here because the data type is unsigned

kisvegabor:
I can't test it now but I believe the changes are tested on your side and seemingly there is nothing harmful in it :slightly_smiling_face: 

So I merge it, thank you!

PR #246, opened 2022-11-03
fix(wayland-protocols): minimum version

gorogoro2:
wayland driver needs configure_bounds and it was introduced on version 1.25: https://github.com/wayland-project/wayland-protocols/commit/344048614ad4a3887e431f63b6f12208edce51f9


kisvegabor:
Thank you!

PR #248, opened 2022-11-04
fix(fbdev): power display on if FBDEV_DISPLAY_POWER_ON is set

agambier:
On some embedded devices the framebuffer driver fails when calling `ioctl(fbfd, FBIOBLANK, FB_BLANK_UNBLANK)` to power the display on.


kisvegabor:
Thank you!

PR #253, opened 2022-11-12
feature(wayland): add double buffering

WallaceIT:
Add double buffering to Wayland driver: one buffer is updated while another one is being displayed.

This doubles the memory requirement, but lowers the possibility to encounter a busy buffer when LVGL core is ready to draw.

Signed-off-by: Francesco Valla <valla.francesco@gmail.com>

### DIFF
--- a/wayland/wayland.c
+++ b/wayland/wayland.c
@@ -119,6 +119,11 @@ struct buffer_hdl
     int size;
     struct wl_buffer *wl_buffer;
     bool busy;
+
+    lv_area_t inv_areas[LV_INV_BUF_SIZE];
+    uint8_t inv_area_joined[LV_INV_BUF_SIZE];
+    uint16_t inv_p;
+    bool is_first_flush;
 };
 
 struct buffer_allocator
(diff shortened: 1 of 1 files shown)


WallaceIT:
Please note that the MR is marked as draft as it is still, in fact, a draft: some of the logic for a proper double buffering is still lacking. Moreover, the entire work will probably be converted to a `direct_mode` implementation for better performance (no data copy).

WallaceIT:
Added carry-over of previous buffer damages, proper double-buffering logic should be in place now.

HR1025:
And the problem https://github.com/lvgl/lv_drivers/pull/242 still exist. (I haven't found the reason so far.)
> you can use `weston --use-pixman` to test on you machine.
Thanks @WallaceIT

HR1025:
Recently, I was studying the implementation of `weston-terminal`. It need some time. I think we can find a good solution from it.
The code is located in `weston/clients/terminal.c` in (weston)[https://gitlab.freedesktop.org/wayland/weston]

Issue #254, opened 2022-11-30
win32drv poor keyboard functionality


wrgallo:
Currently the `win32drv` keyboard driver wouldn't support lots of keys, such as `"+-*/[];',./\!@#$%^&*()_+~"`, also the letters `"A-Z"` would always become lower-case regardless of the current state of `shift` and `caps-lock` keys.

I am already working on the pull-request to fix this issue.

MouriNaruto:
@kisvegabor It's bug I can reproduce. I will fix it ASAP.

Kenji Mouri

MouriNaruto:
@wrgallo Your issue have been fixed. I will create a PR.

![image](https://user-images.githubusercontent.com/10867563/205047767-84b75987-5084-422c-ab77-12b5d26220f1.png)

Kenji Mouri

Issue #256, opened 2022-11-30
win32drv Possibility to change the window title for each display


wrgallo:
Currently by creating a display window with `lv_win32_init` the created window will have the title `L"LVGL Simulator for Windows Desktop"`, that string cannot be easily changed by user.

I am already working on a pull request to fix this issue.

kisvegabor:
@MouriNaruto could you take look at it?

kisvegabor:
Thank you!

MouriNaruto:
@kisvegabor 

The issue is also solved by https://github.com/lvgl/lv_drivers/pull/259.

Kenji Mouri

kisvegabor:
Great, I close it then.

Issue #255, opened 2022-11-30
win32drv has no support for multiple displays


wrgallo:
Currently the `win32drv` only support a single display, looking at the source-code there are lots of static variables to control that specific display window.
```C
static HWND g_window_handle = NULL;
static HDC g_buffer_dc_handle = NULL;
static UINT32* g_pixel_buffer = NULL;
static SIZE_T g_pixel_buffer_size = 0;
static lv_disp_t* g_display = NULL;
static bool volatile g_display_refreshing = false;
static bool volatile g_mouse_pressed = false;
static LPARAM volatile g_mouse_value = 0;
static bool volatile g_mousewheel_pressed = false;
static int16_t volatile g_mousewheel_value = 0;
static bool volatile g_keyboard_pressed = false;
static WPARAM volatile g_keyboard_value = 0;
```

Also there are 3 global variables to windows drivers input devices (mouse , keyboard and mousewheel).
```C
EXTERN_C lv_indev_t* lv_win32_pointer_device_object;
EXTERN_C lv_indev_t* lv_win32_keypad_device_object;
EXTERN_C lv_indev_t* lv_win32_encoder_device_object;
```
But, each input driver needs to be assigned to a display which makes those global variable useless for other displays.

I am already working on a pull request to fix this issue.

kisvegabor:
cc @MouriNaruto

MouriNaruto:
@kisvegabor Got it. I will try to add this feature to win32drv.

Kenji Mouri

kisvegabor:
Thank you!

MouriNaruto:
@wrgallo @kisvegabor 

I have just added the multiple display support and preparing the PR.

![image](https://user-images.githubusercontent.com/10867563/206708247-487fedc2-71b8-4d5f-84a5-b2a0ca815bf4.png)

Kenji Mouri

PR #257, opened 2022-11-30
refactor(win32drv): multiple display support

wrgallo-navitas:
Fixes: #254 #255 #256

Code-reviews are welcome

Co-Authored-By: [wrgallo](https://github.com/wrgallo)

### DIFF
--- a/win32drv/win32drv.c
+++ b/win32drv/win32drv.c
@@ -15,13 +15,13 @@
 #include <process.h>
 #include <stdbool.h>
 #include <stdint.h>
+#include <stdio.h>
 
 /*********************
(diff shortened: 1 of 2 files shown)


MouriNaruto:
I don‘t think having maximum display count limit is acceptable. Also, the window class should be reused as possible because windows ATOM object have maximum count limit.

For keyboard support improvement, I think https://github.com/lvgl/lv_drivers/pull/258 is simpler than this PR and add mutex for improve the reliability.

As an experienced Windows C++ application developer, I want to provide a better solution than this PR.

Kenji Mouri

Issue #267, opened 2023-01-23
SDL + multiple display + touch is not working


Viatorus:
The support for touch events for SDL was added with: https://github.com/lvgl/lv_drivers/pull/97

The problem:
We are using multiple lvgl display drivers which all are visible at the same time e.g.:

![image](https://user-images.githubusercontent.com/3637911/214008117-c09bd856-fda0-48ea-8c2a-ea1f9d16b547.png)

The inner display is our "most active" display.

But with `last_x = LV_HOR_RES * event->tfinger.x / MONITOR_ZOOM;` the touch coordinates are wrong.

I am not quit sure how to fix this at the moment. :/

Viatorus:
I think the correct way would be to usage of SDL_HOR_RES and SDL_VER_RES. I think this was forgotten by refactoring.

PR #268, opened 2023-01-23
fix: use SDL_RES instead of LV_RES for touch events

Viatorus:
Fixes: https://github.com/lvgl/lv_drivers/issues/267

@stawnPTL @embeddedt


kisvegabor:
We should really use `SDL_HOR/VER_RES` here.

Thank you!

PR #270, opened 2023-02-02
feature(wayland): Add a shared memory manager.

simplejack-src:
Allowing backing buffers to be allocated on demand. In addition, the shared memory manager tracks buffer acquisition history (in Wayland EGL parlance 'buffer age'), allowing the acquiring code to perform a delta initialization (apply changes from previously acquired buffers to the most recent, as oppose to a full buffer initialization).
In particular, the shared memory manager, maintains a shared memory pool, wherein individual buffers, belonging to separate buffer groups (which all have the same size) are allocated from. A LRU cache is then used on top of this to keep a hot cycle (i.e. swap chain) available. Finally, the acquiring code, caches individual buffer damage in a ring buffer to be able to apply the aforementioned delta initialization.

This PR is my completion of the work started in #250.

Notes:
- There's some toggles implemented (namely the ring buffer size and shared memory manager preferred number of buffers), which can be changed to tradeoff swap-chain length (and thus delta initialization time) and allocation overhead.
- Still a lot of places where performance could be improved (and heuristics added), in particular the ring buffer could use a weighted size to make a decision of when to delta initialize or full initialize a buffer
- This implementation has mostly been hacked on to the existing Wayland driver, to hopefully fix-up a couple bugs/make it more Wayland conformant
- Moving forward I may look to implement the SHM interface backend on SDL, as I believe with this the SDL driver would be feature superior to this driver

### DIFF
--- a/wayland/smm.c
+++ b/wayland/smm.c
@@ -1,0 +1,648 @@
+/**
+ * @file smm.c
+ *
+ */
+#include <stddef.h>
+#include <errno.h>
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <time.h>
+#include <unistd.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include "smm.h"
+
+#define MAX_NAME_ATTEMPTS (5)
+#define PREFER_NUM_BUFFERS (3)
+
+#define ROUND_UP(n, b) (((((n) ? (n) : 1) + (b) - 1) / (b)) * (b))
+#define LLHEAD(type) \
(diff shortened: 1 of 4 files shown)


kisvegabor:
Thank you very much @simplejack-src. Can I merge it or are we waiting feedback from someone?

simplejack-src:
It works for me(TM) :). @WallaceIT see anything out of place?

WallaceIT:
Seems very well done! I don't have time nor means to test it since I'm travelling, though.

kisvegabor:
Thanks for taking a look @WallaceIT.

I'll merge it now then. Please open a new issue if you find an issues later.

PR #273, opened 2023-02-17
Use win32 singly linked list instead of LVGL linked list for fixing memory access conflict issues for win32drv.

MouriNaruto:
This issue is reported by @SmartAnda.

image

Memory access conflicts occur when frequent keyboard navigation is performed due to LVGL's implementation does not thread safe. 

So, use win32 singly linked list instead of LVGL linked list for solving the issue without preparing a modified version of LVGL linked list implementation and keeping the win32drv implementation as simple as possible.

Kenji Mouri

### DIFF
--- a/win32drv/win32drv.c
+++ b/win32drv/win32drv.c
@@ -12,6 +12,7 @@
 #if USE_WIN32DRV
 
 #include <windowsx.h>
+#include <malloc.h>
 #include <process.h>
 #include <stdbool.h>
 #include <stdint.h>
@@ -165,6 +166,25 @@ static LRESULT CALLBACK lv_win32_window_message_callback(
 
 static unsigned int __stdcall lv_win32_window_thread_entrypoint(
     void* raw_parameter);
+
+static void lv_win32_push_key_to_keyboard_queue(
+    lv_win32_window_context_t* context,
+    uint32_t key,
+    lv_indev_state_t state)
+{
+    lv_win32_keyboard_queue_item_t* current =
+        (lv_win32_keyboard_queue_item_t*)(_aligned_malloc(
+            sizeof(lv_win32_keyboard_queue_item_t),
+            MEMORY_ALLOCATION_ALIGNMENT));
+    if (current)
+    {
+        current->key = key;
+        current->state = state;
+        InterlockedPushEntrySList(
+            context->keyboard_queue,
+            &current->ItemEntry);
+    }
+}
 
 /**********************
  *  GLOBAL VARIABLES
@@ -710,14 +730,14 @@ static void lv_win32_keypad_driver_read_callback(
     EnterCriticalSection(&context->keyboard_mutex);
 
     lv_win32_keyboard_queue_item_t* current =
-        (lv_win32_keyboard_queue_item_t*)(_lv_ll_get_head(
-            &context->keyboard_queue));
+        (lv_win32_keyboard_queue_item_t*)(InterlockedPopEntrySList(
+            context->keyboard_queue));
     if (current)
     {
         data->key = current->key;
         data->state = current->state;
 
-        _lv_ll_remove(&context->keyboard_queue, current);
+        _aligned_free(current);
 
         data->continue_reading = true;
     }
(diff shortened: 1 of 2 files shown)


kisvegabor:
Looks good, thank you!

MaD70:
@kisvegabor I don't understand how this can work: at first I thought that the proposed patch was accepted by mistake, because the data structure in [interlockedapi.h](https://learn.microsoft.com/en-us/windows/win32/api/interlockedapi/) is a **stack** not a queue ([`InterlockedPopEntrySList`](https://github.com/lvgl/lv_drivers/pull/273/files#diff-55553580a0fd661c43bbc7d2e78a6d502993b7b91d4f28090dd686b63af40da9R733) and [`InterlockedPushEntrySList`](https://github.com/lvgl/lv_drivers/pull/273/files#diff-55553580a0fd661c43bbc7d2e78a6d502993b7b91d4f28090dd686b63af40da9R183), used in the patch, _removes/inserts an item from the **front** of a singly linked list_).

Then I noticed that indeed the code that the patch replaced uses the `lv_ll_t` data structure in `lv_ll.h` as a **stack** (pushes and pops from head with [`_lv_ll_ins_head`](https://github.com/lvgl/lv_drivers/pull/273/files#diff-55553580a0fd661c43bbc7d2e78a6d502993b7b91d4f28090dd686b63af40da9L1031) and [`_lv_ll_get_head`](https://github.com/lvgl/lv_drivers/pull/273/files#diff-55553580a0fd661c43bbc7d2e78a6d502993b7b91d4f28090dd686b63af40da9L713)) and not as a queue. So at least the name of the [`keyboard_queue`](https://github.com/lvgl/lv_drivers/pull/273/files#diff-7454e06a6de01dd69bfc7b215908652b4edcc017261d2ff779e0b3daed843da7L90) variable is misleading, but I wonder what's the point of processing the sequence of keystrokes always starting from the last one and going backwards (stack).

MouriNaruto:
@MaD70 

> I don't understand how this can work.

Because LVGL's linked list doesn't support thread safe. And win32drv uses separate window thread to process the window message. (Make main thread can have more time to do with LVGL's scheduler.)

> So at least the name of the [keyboard_queue](https://github.com/lvgl/lv_drivers/pull/273/files#diff-7454e06a6de01dd69bfc7b215908652b4edcc017261d2ff779e0b3daed843da7L90) variable is misleading

Not misleading, because it's preprocessed Win32's keyboard input queue.

Kenji Mouri

MaD70:
> Because LVGL's linked list doesn't support thread safe. And win32drv uses separate window thread to process the window message. (Make main thread can have more time to do with LVGL's scheduler.)

@MouriNaruto that's not the part that is not clear to me (I linked to [interlockedapi.h](https://learn.microsoft.com/en-us/windows/win32/api/interlockedapi/), I understood perfectly that the patch was made to make treadsafe the processing of keys from keyboard).


> Not misleading, because it's preprocessed Win32's keyboard input queue.

I see that the keys in `keyboard_queue` **come** from the Win32's keyboard input queue (the [`lv_win32_window_message_callback`](https://github.com/lvgl/lv_drivers/blob/master/win32drv/win32drv.c#L775) event handler with the classic switch to dispatch events like [the case for `WM_KEYDOWN` and `WM_KEYUP`](https://github.com/lvgl/lv_drivers/blob/master/win32drv/win32drv.c#L997)) but `keyboard_queue` itself is created and managed by LVGL, it's not the Win32's keyboard input queue. I'm asking why `keyboard_queue` is managed like a stack.

MaD70:
> For each time LVGL keyboard processing reads, it is like the queue.

Ok, I trust you on this (to understand why this is true I would have to study the library in much more detail, it's not at all apparent from a normal reading and that's why I asked).

> Choose stack because we don't want to use pure C (because LVGL is a project as pure C as possible) to implement the thread safe queue by ourselves which is not tested by history.

I understand this, concurrency is notoriously difficult to implement correctly. I'm not aware of any current and widely used C implementation of a thread safe queue.

If the need arises, I would suggest the algorithms adopted by the Java library [Simple, Fast, and Practical Non-Blocking and Blocking Concurrent Queue Algorithms](https://www.cs.rochester.edu/research/synchronization/pseudocode/queues.html): on this web page there is pseudo-code for the implementation and [at this link there is an old C implementation](http://www.cs.rochester.edu/research/synchronization/code/concurrent_queues/SGI.tgz) (.tgz) (accessed from this page: [High-Performance Synchronization for Shared-Memory Parallel Programs](https://cs.rochester.edu/research/systems/high_performance_synch/index.html), the link is in the section _Legacy Executables_ at _Fast concurrent queue algorithms_).

The implementation should be modernized to use [C11 Atomic operations library](https://en.cppreference.com/w/c/atomic), in particular [`CAS` (Compare-And-Swap)](https://en.wikipedia.org/wiki/Compare-and-swap) in pseudo-code should be one of the [`atomic_compare_exchange`](https://en.cppreference.com/w/c/atomic/atomic_compare_exchange) of C11.

PR #274, opened 2023-03-10
feature(libinput): Add function to reset driver states

Johennes:
This adds new functions to de-initialise the libinput and XKB driver states which allows existing data structures to be reused when devices are connected or disconnected at runtime.

An example of how to use this can be found in https://gitlab.com/cherrypicker/unl0kr/-/merge_requests/15.

### DIFF
--- a/indev/libinput.c
+++ b/indev/libinput.c
@@ -138,8 +138,8 @@ bool libinput_set_file_state(libinput_drv_state_t *state, char* dev_name)
   // citing libinput.h:libinput_path_remove_device:
   // > If no matching device exists, this function does nothing.
   if (state->libinput_device) {
-    state->libinput_device = libinput_device_unref(state->libinput_device);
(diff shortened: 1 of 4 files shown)


indev/libinput.c
```diff
@@ -138,8 +138,8 @@ bool libinput_set_file_state(libinput_drv_state_t *state, char* dev_name)
   // citing libinput.h:libinput_path_remove_device:
   // > If no matching device exists, this function does nothing.
   if (state->libinput_device) {
-    state->libinput_device = libinput_device_unref(state->libinput_device);
     libinput_path_remove_device(state->libinput_device);
+    state->libinput_device = libinput_device_unref(state->libinput_device);
```
Johennes, line 142: I changed this while at it because this seemed like the more logical / safe order.

kisvegabor:
Thank you!

PR #275, opened 2023-03-10
feature(libinput): Expose function for querying capability

Johennes:
This extracts the code for querying input device capabilities into a global function so that consumers of the library can use it as well.

An example of how this is used can be found in https://gitlab.com/cherrypicker/unl0kr/-/merge_requests/15.


kisvegabor:
I agree that it'd be useful. Thanks!

PR #276, opened 2023-03-13
fix(xkb): Fix memory leak by avoiding double reference

Johennes:
This leak was reported by Valgrind in unl0kr (see https://gitlab.com/cherrypicker/unl0kr/-/issues/43). It turns out that `xkb_keymap_new` and `xkb_state_new` return referenced instances. So the extra calls to `xkb_keymap_ref` and `xkb_state_ref` were not needed and caused the memory to not be properly cleaned up in `xkb_deinit_state`.

Apologies for introducing this bug together with the original XKB work in lv_drivers. I was mistakenly assuming this was the correct way to use the API because the libinput API worked this way.

### DIFF
--- a/indev/xkb.c
+++ b/indev/xkb.c
@@ -79,14 +79,14 @@ bool xkb_init_state(xkb_drv_state_t *state) {
  * @param state XKB driver state to use
  */
 void xkb_deinit_state(xkb_drv_state_t *state) {
-  if (state->keymap) {
-    xkb_keymap_unref(state->keymap);
-    state->keymap = NULL;
-  }
-
   if (state->state) {
     xkb_state_unref(state->state);
     state->state = NULL;
+  }
+
+  if (state->keymap) {
+    xkb_keymap_unref(state->keymap);
+    state->keymap = NULL;
   }
 }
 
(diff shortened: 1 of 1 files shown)


indev/xkb.c
```diff
@@ -79,15 +79,15 @@ bool xkb_init_state(xkb_drv_state_t *state) {
...
-
   if (state->state) {
     xkb_state_unref(state->state);
     state->state = NULL;
   }
+
+  if (state->keymap) {
+    xkb_keymap_unref(state->keymap);
+    state->keymap = NULL;
+  }
```
Johennes, line 90: This shouldn't really matter much but the state references the keymap. So it seems more logical to free the memory in this order.

kisvegabor:
Thanks for the fix! :+1:

kisvegabor:
BTW, now LVGL `master` has the new architecture and we can add the drivers from this repo. This way we'd have a lot of drivers in a central place, and they would be versioned and maintained together. 

You you be interested in contributing by moving this driver to LVGL `master`? [Here](https://github.com/lvgl/lvgl/issues/4011) is a summary about the changes.

Johennes:
> BTW, now LVGL `master` has the new architecture and we can add the drivers from this repo. This way we'd have a lot of drivers in a central place, and they would be versioned and maintained together.
> 
> You you be interested in contributing by moving this driver to LVGL `master`? [Here](https://github.com/lvgl/lvgl/issues/4011) is a summary about the changes

Oh, great! Having them in a single repository would also make the packaging a bit easier for us. We're currently investigating keys being dropped when typing too fast with the current libinput driver (https://gitlab.com/cherrypicker/unl0kr/-/issues/36). Looks like switching to the new driver architecture on LVGL `master` would be a good next step before we dive deeper into this. :+1:

kisvegabor:
Happy to hear that you are interested in it! :relaxed: 

Let me know when you'd like to jump into it and need some assistance.

PR #277, opened 2023-04-04
Add a way to set SDL window title from lv_drv_conf.h

hanhsuan:
I'm trying to use LVGL and SDL to write one cross platform GUI application, but the window title for SDL is hardcode in the sdl.c. Therefore, it would be nice to have a way to set SDL window title from lv_drv_conf.h just as other setting.

### DIFF
--- a/lv_drv_conf_template.h
+++ b/lv_drv_conf_template.h
@@ -110,6 +110,9 @@
 
 /*Open two windows to test multi display support*/
 #  define SDL_DUAL_DISPLAY            0
+
+/* Window Title  */
+#  define SDL_WINDOW_TITLE "TFT Simulator"
 #endif
 
 /*-------------------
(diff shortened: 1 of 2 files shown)


kisvegabor:
It's ok, but please add an 
```c
#ifndef SDL_WINDOW_TITLE
#define SDL_WINDOW_TITLE "TFT Simulator"
#endif
```

in `sdl.c`.

Note that, we have already moved the SDL driver to [LVGL master](https://github.com/lvgl/lvgl/tree/master/src/dev/sdl) branch. When LVGL v9 will be released this repository will be detracted. So it might be worth adding the same feature in LVGL master too.

hanhsuan:
I will go to add this feature to the LVGL master. Thanks a lot for your notice.

hanhsuan:
Should I follow what lvgl v9 to modify this PR, or close ?

kisvegabor:
Sorry for the delay.

This repo mostly uses `define` based configurations we can keep it as it is.

Merging

PR #280, opened 2023-06-04
Libinput and fbdev improvements for unl0kr

kcxt:
Add a framebuffer quirk to force the display to refresh after every draw, this is needed for some quirky panels and DRM/fbdev emulation setups. It requires that the display buffer is set to the same size as the whole framebuffer so that the entire display can be flushed at once if needed, otherwise there is tearing.

Also use pthreads and a small circular queue to stop the libinput driver from dropping inputs. The LVGL tickrate can't always keep up and this ensures that inputs aren't dropped.

The keypad support needs to be tested.

Cc: @Johennes

### DIFF
--- a/indev/libinput.c
+++ b/indev/libinput.c
@@ -541,21 +541,21 @@ static void read_pointer(libinput_drv_state_t *state, struct libinput_event *eve
       break;
     }
     case LIBINPUT_EVENT_TOUCH_UP:
+      /*
+       * We don't support "multitouch", but libinput does. To make fast typing with two thumbs
+       * on a keyboard feel good, it's necessary to handle two fingers individually. The edge
+       * case here is if you press a key with one finger and then press a second key with another
+       * finger. No matter which finger you release, it will count as the second finger releasing
+       * and ignore the first because LVGL only stores a single (the latest) pressed state.
+       *
+       * To work around this, we detect the case where one finger is released while the other is
+       * still pressed and insert dummy events so that both release events trigger at the correct
+       * position.
+       */
       if (slot == 0 && state->slots[1].pressed == LV_INDEV_STATE_PR) {
-        /*
-        * We don't support "multitouch", but libinput does. To make fast typing with two thumbs
-        * on a keyboard feel good, it's necessary to handle two fingers individually. The edge
-        * case here is if you press a key with one finger and then press a second key with another
-        * finger. No matter which finger you release, it will count as the second finger releasing
-        * and ignore the first because LVGL only stores a single (the latest) pressed state.
-        *
-        * To work around this, detect the case where the first finger is releasing while the second is
-        * still pressed and insert dummy press events for the first and second finger around the
-        * release event for the first finger.
-        *
-        * In other words, P1 > P2 > R1 > R2 becomes P1 > P2 > (P1) > R1 > (P2) > R2 where P and R
-        * mean press and release, respectively, and the parentheses denote the dummy events.
-        */
+        /* The first finger is released while the second finger is still pressed.
+         * We turn P1 > P2 > R1 > R2 into P1 > P2 > (P1) > R1 > (P2) > R2.
+         */
 
         /* Inject the dummy press event for the first finger */
         libinput_lv_event_t *synth_evt = evt;
@@ -570,6 +570,18 @@ static void read_pointer(libinput_drv_state_t *state, struct libinput_event *eve
         synth_evt = new_event(state);
         synth_evt->pressed = LV_INDEV_STATE_PR;
         synth_evt->point = state->slots[1].point;
+      } else if (slot == 1 && state->slots[0].pressed == LV_INDEV_STATE_PR) {
+        /* The second finger is released while the first finger is still pressed.
+         * We turn P1 > P2 > R2 > R1 into P1 > P2 > R2 > (P1) > R1.
+         */
+
+        /* Append the real release event for the second finger */
+        evt->pressed = LV_INDEV_STATE_REL;
+
+        /* Inject the dummy press event for the first finger */
+        libinput_lv_event_t *synth_evt = new_event(state);
+        synth_evt->pressed = LV_INDEV_STATE_PR;
+        synth_evt->point = state->slots[0].point;
       } else {
         evt->pressed = LV_INDEV_STATE_REL;
       }
--- a/indev/libinput.c
+++ b/indev/libinput.c
@@ -363,14 +363,26 @@ void libinput_read_state(libinput_drv_state_t * state, lv_indev_drv_t * indev_dr
   libinput_lv_event_t *evt = get_event(state);
   data->continue_reading = event_pending(state);
   if (!evt)
-    evt = &state->last_event;
-  /* indev expects us to report the most recent state */
-  else if (!data->continue_reading)
-    state->last_event = *evt;
+    evt = &state->last_event; /* indev expects us to report the most recent state */
+
+  /* Rewrite relative pointer click and motion events into absolute ones using the previous event */
+  if (evt->is_relative) {
+    /* We need to read unrotated display dimensions directly from the driver because libinput won't account
+     * for any rotation inside of LVGL */
+    lv_disp_drv_t *drv = lv_disp_get_default()->driver;
+
+    evt->point.x = state->last_event.point.x + evt->point.x;
+    evt->point.y = state->last_event.point.y + evt->point.y;
+    evt->point.x = LV_CLAMP(0, evt->point.x, drv->hor_res - 1);
+    evt->point.y = LV_CLAMP(0, evt->point.y, drv->ver_res - 1);
+    evt->is_relative = false;
+  }
 
   data->point = evt->point;
   data->state = evt->pressed;
   data->key = evt->key_val;
+
+  state->last_event = *evt; /* Remember the last event for the next call */
 
   pthread_mutex_unlock(&state->event_lock);
 
@@ -545,6 +557,7 @@ static void read_pointer(libinput_drv_state_t *state, struct libinput_event *eve
       evt->point.y += libinput_event_pointer_get_dy(pointer_event);
       evt->point.x = LV_CLAMP(0, evt->point.x, drv->hor_res - 1);
       evt->point.y = LV_CLAMP(0, evt->point.y, drv->ver_res - 1);
+      evt->is_relative = true;
       break;
     case LIBINPUT_EVENT_POINTER_MOTION_ABSOLUTE:
       pointer_event = libinput_event_get_pointer_event(event);
@@ -559,6 +572,7 @@ static void read_pointer(libinput_drv_state_t *state, struct libinput_event *eve
     case LIBINPUT_EVENT_POINTER_BUTTON:
       enum libinput_button_state button_state = libinput_event_pointer_get_button_state(pointer_event); 
       evt->pressed = button_state == LIBINPUT_BUTTON_STATE_RELEASED ? LV_INDEV_STATE_REL : LV_INDEV_STATE_PR;
+      evt->is_relative = true;
       break;
     default:
       break;
--- a/indev/libinput.c
+++ b/indev/libinput.c
@@ -284,15 +284,6 @@ libinput_lv_event_t *get_event(libinput_drv_state_t *state)
   return evt;
 }
 
-libinput_lv_event_t *peek_event(libinput_drv_state_t *state)
-{
-  if (state->start == state->end) {
-    return NULL;
-  }
-
-  return &state->points[state->start];
-}
-
 bool event_pending(libinput_drv_state_t *state)
 {
   return state->start != state->end;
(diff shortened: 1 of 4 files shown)


indev/libinput.c
```diff
@@ -250,6 +267,83 @@ void libinput_read(lv_indev_drv_t * indev_drv, lv_indev_data_t * data)
   libinput_read_state(&default_state, indev_drv, data);
 }
 
+libinput_lv_event_t *new_event(libinput_drv_state_t *state)
+{
+  libinput_lv_event_t *evt = &state->points[state->end];
+
+  if (++state->end == MAX_EVENTS)
```
Johennes, line 274: Would you also have to do something to `state->start` here to prevent `end` overtaking `start` which would then read later events first when accessing `start`?

indev/libinput.c
```diff
@@ -259,37 +353,24 @@ void libinput_read(lv_indev_drv_t * indev_drv, lv_indev_data_t * data)
...
-  rc = poll(state->fds, nfds, timeout);
-  switch (rc){
-    case -1:
-      perror(NULL);
-    case 0:
-      goto report_most_recent_state;
-    default:
-      break;
+  pthread_mutex_lock(&state->event_lock);
+  if (!event_pending(state)) {
```
Johennes, line 359: I think you can remove this `if` because `get_event` does the same test and returns `NULL` if it fails which is then skipped by the `if (evt)` below and ends up in the same unlock and return behavior.

indev/libinput.c
```diff
@@ -410,39 +491,51 @@ static void reset_scanned_devices(void) {
...
-      state->most_recent_touch_point.y = y;
-      state->button = LV_INDEV_STATE_PR;
+      evt->point.x = x;
+      evt->point.y = y;
+      evt->pressed = LV_INDEV_STATE_PR;
       break;
     case LIBINPUT_EVENT_TOUCH_UP:
-      state->button = LV_INDEV_STATE_REL;
+      evt = new_event(state);
+      touch_event = libinput_event_get_touch_event(event);
```
Johennes, line 522: I think this isn't actually used anymore because the line below is commented out?

Johennes (commented):
Pretty amazing stuff! Thanks @calebccff!

I've only given it a quick look this morning. Will try to do some testing against https://gitlab.com/cherrypicker/unl0kr/-/merge_requests/21 this week.

@kisvegabor, I know, lv_drivers is supposed to be deprecated in future. It'll be very helpful to have these changes merged though because https://gitlab.com/cherrypicker/unl0kr still depends on it.

We already started migrating to lvgl `master` in https://gitlab.com/cherrypicker/unl0kr/-/merge_requests/20 and we'll port the changes to the fbdev driver and the libinput driver, too, as part of this.

kisvegabor (commented):
Really nice, thank you!

We can merge it here, but yeah... it'd be great to add these updates to master as well. :slightly_smiling_face:

indev/libinput.c
```diff
@@ -259,37 +353,25 @@ void libinput_read(lv_indev_drv_t * indev_drv, lv_indev_data_t * data)
...
+  libinput_lv_event_t *evt = get_event(state);
+  data->continue_reading = event_pending(state);
+  if (!evt)
+    evt = &state->last_event;
+  /* indev expects us to report the most recent state */
+  else if (!data->continue_reading)
+    state->last_event = *evt;
+
+  data->point = evt->point;
+  data->state = evt->pressed;
```
Johennes, line 369: I think this is missing `data->key = evt->key_val`?

indev/libinput.c
```diff
@@ -457,50 +554,51 @@ static void read_pointer(libinput_drv_state_t *state, struct libinput_event *eve
 static void read_keypad(libinput_drv_state_t *state, struct libinput_event *event) {
   struct libinput_event_keyboard *keyboard_event = NULL;
   enum libinput_event_type type = libinput_event_get_type(event);
+  libinput_lv_event_t *evt = NULL;
```
Johennes, line 557: This doesn't appear to ever be set to something else than `NULL`. I think you're missing an `evt = new_event(state);`  inside the `case LIBINPUT_EVENT_KEYBOARD_KEY` below?

indev/libinput.c
```diff
@@ -419,30 +502,44 @@ static void read_pointer(libinput_drv_state_t *state, struct libinput_event *eve
   switch (type) {
     case LIBINPUT_EVENT_TOUCH_MOTION:
     case LIBINPUT_EVENT_TOUCH_DOWN:
+    case LIBINPUT_EVENT_TOUCH_UP:
       touch_event = libinput_event_get_touch_event(event);
+      /* ignore more than 2 fingers to avoid overflowing the buffer */
+      if (libinput_event_touch_get_slot(touch_event) > 1)
+        return;
+      evt = new_event(state);
```
Johennes, line 510: I think something has went missing here. `evt` is only set here for `LIBINPUT_EVENT_TOUCH_UP` but it's written into in all the other cases below?

calebccff: The other cases here fall through, I split this out as "common handling code", I'll rewrite it to be a bit clearer.

calebccff: oh I missed the LIBINPUT_EVENT_POINTER_* cases too... nice catch :sweat_smile:

calebccff:
Totally forgot to push :sweat_smile: sorry for the wait. The touch/pointer handling was actually pretty broken... Anyways, hopefully this is good to go now!

Johennes:
Thanks for updating the PR @calebccff and sorry for the delay. I will take a look at this and test it within unl0kr but I won't manage before next week, unfortunately.

indev/libinput.c
```diff
@@ -186,6 +185,7 @@ bool libinput_set_file_state(libinput_drv_state_t *state, char* dev_name)
  */
 void libinput_init(void)
 {
+  memset(&default_state, 0, sizeof(libinput_drv_state_t));
```
Johennes, line 188: Doesn't this undo `default_state = { .event_lock = PTHREAD_MUTEX_INITIALIZER, };`?

I'm also not seeing any call to `pthread_mutex_init` in the corresponding [unl0kr MR](https://gitlab.com/cherrypicker/unl0kr/-/merge_requests/21/diffs). Could it be that we're accidentally using `event_lock` from `default_state` in the dynamic states in unl0kr? They are cleared with `lv_memzero` so `event_lock` should always result in `0` which might be what `PTHREAD_MUTEX_INITIALIZER` returns in this case.

I think it would actually be good to call `pthread_mutex_init` inisde of `libinput_init_state` because the caller shouldn't have to know anything about the thread handling inside of the driver. You may just have to make sure that `libinput_init` doesn't initialize the lock in `default_state` twice by calling `libinput_init_state` but not sure if that would actually be a problem

indev/libinput.c
```diff
@@ -410,39 +493,55 @@ static void reset_scanned_devices(void) {
...
+      type != LIBINPUT_EVENT_POINTER_MOTION &&
+      type != LIBINPUT_EVENT_POINTER_BUTTON)
+    return;
+
   /* We need to read unrotated display dimensions directly from the driver because libinput won't account
    * for any rotation inside of LVGL */
   lv_disp_drv_t *drv = lv_disp_get_default()->driver;
 
+  /* ignore more than 2 fingers as it will only confuse LVGL */
+  touch_event = libinput_event_get_touch_event(event);
```
Johennes, line 512: This triggers a warning on the console when the event is one of the `LIBINPUT_EVENT_POINTER_*` types. So for mice and trackpads this prints on every cursor motion or key press.

indev/libinput.c
```diff
@@ -410,39 +493,55 @@ static void reset_scanned_devices(void) {
...
+      evt->pressed = LV_INDEV_STATE_REL;
       break;
     case LIBINPUT_EVENT_POINTER_MOTION:
       pointer_event = libinput_event_get_pointer_event(event);
-      state->most_recent_touch_point.x += libinput_event_pointer_get_dx(pointer_event);
-      state->most_recent_touch_point.y += libinput_event_pointer_get_dy(pointer_event);
-      state->most_recent_touch_point.x = LV_CLAMP(0, state->most_recent_touch_point.x, drv->hor_res - 1);
-      state->most_recent_touch_point.y = LV_CLAMP(0, state->most_recent_touch_point.y, drv->ver_res - 1);
+      evt->point.x += libinput_event_pointer_get_dx(pointer_event);
+      evt->point.y += libinput_event_pointer_get_dy(pointer_event);
```
Johennes, line 537: Upon testing with a trackpad, it doesn't let me move the cursor from the top left corner of the screen and always jumps back. I think it's caused by this part here. We used to add the delta from `libinput_event_pointer_get_dx/y` to the most recently read coordinates before. But here `evt` is a blank slot that we've zero'ed out so we're always starting in the top left corner regardless of where the cursor moved before.

We could look back in the ring buffer and add the value of `point` from the previous event. However, we'd also have to account for overrunning the two pointers in the buffer again and I'm not sure what value we'd set for `point` in that case. What might be better instead is to store `dx` and `dy` here and then evaluate either `x` and `y` or `dx` and `dy` in `libinput_read_state` by making use of `last_event`. I think we'd also have to change that function to always update `last_event` though regardless of whether more events are pending or not.

Johennes (changes requested):
Sorry again for the delay. I reviewed the code and tested it with https://gitlab.com/cherrypicker/unl0kr/-/merge_requests/21 in unl0kr.

Pointer devices don't currently work correctly because their delta motion logic needs special handling with the new event buffer.

There is possibly also a problem with the mutex. I didn't notice any issues related to that when testing but I think the mutex initialization is missing and might just work accidentally in unl0kr.

indev/libinput.c
Johennes: Proposal how to fix this: https://github.com/calebccff/lv_drivers/pull/1

indev/libinput.c
Johennes: Proposal for how to fix this: https://github.com/calebccff/lv_drivers/pull/2

indev/libinput.c
```diff
@@ -211,6 +211,8 @@ void libinput_init_state(libinput_drv_state_t *state, char* path)
   state->fds[0].events = POLLIN;
   state->fds[0].revents = 0;
 
+  pthread_create(&state->worker_thread, NULL, libinput_poll_worker, state);
```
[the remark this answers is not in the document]
Johennes: I included this in https://github.com/calebccff/lv_drivers/pull/2

indev/libinput.c
Johennes: Proposal how to fix this: https://github.com/calebccff/lv_drivers/pull/3

calebccff:
Rebased on master and picked up Johannes' fixes (Thanks a lot for that!). Should be good to go! Tested again with unl0kr.

I added a new commit to properly handle typing with two thumbs by injecting some fake events, with this you can type as fast as you want with your thumbs and not miss an input which is really nice for unl0kr. See the commit description for more details.

I get that this is a bit of a hack, I'm happy to drop it if it blocks the rest of the series.

indev/libinput.c
```diff
@@ -410,50 +550,79 @@ static void reset_scanned_devices(void) {
...
     case LIBINPUT_EVENT_POINTER_MOTION:
-      pointer_event = libinput_event_get_pointer_event(event);
-      state->most_recent_touch_point.x += libinput_event_pointer_get_dx(pointer_event);
-      state->most_recent_touch_point.y += libinput_event_pointer_get_dy(pointer_event);
-      state->most_recent_touch_point.x = LV_CLAMP(0, state->most_recent_touch_point.x, drv->hor_res - 1);
-      state->most_recent_touch_point.y = LV_CLAMP(0, state->most_recent_touch_point.y, drv->ver_res - 1);
+      evt->point.x += libinput_event_pointer_get_dx(pointer_event);
+      evt->point.y += libinput_event_pointer_get_dy(pointer_event);
+      evt->point.x = LV_CLAMP(0, evt->point.x, drv->hor_res - 1);
+      evt->point.y = LV_CLAMP(0, evt->point.y, drv->ver_res - 1);
```
Johennes, line 607: I think something may have gone wrong in the cherry picking. This part looks different from https://github.com/calebccff/lv_drivers/pull/3/files. The clamping and the use of `+=` shouldn't be here anymore as they've moved to `libinput_read_state` above. When testing the current branch with unl0kr and a track pad, I can only move the cursor down, not up anymore and I think this is the reason.

indev/libinput.c
```diff
@@ -259,37 +365,71 @@ void libinput_read(lv_indev_drv_t * indev_drv, lv_indev_data_t * data)
...
+    state->slots[1].pressed = LV_INDEV_STATE_REL;
+  } else if (state->doing_mtouch_dummy_event == 1) {
+    /* Now that the position is definitely correct, send the release */
+    evt->pressed = LV_INDEV_STATE_REL;
+    state->doing_mtouch_dummy_event++;
+  } else if (state->doing_mtouch_dummy_event == 2) {
+    /* Finally, update the position to the remaining finger and send a press */
+    evt->pressed = LV_INDEV_STATE_PR;
+    evt->point = state->slots[1].point; /* Wherever slot 1 most recently pressed */
+    state->doing_mtouch_dummy_event = 0;
```
Johennes, line 416: I think you can consume the event here? If I'm reading the code correctly, it'll otherwise be hit again on the next call of this function in which it should enter the `else` right below. It's probably a no-op because it's just triggering the same press event twice but probably worth avoiding.

Johennes:
> Rebased on master and picked up Johannes' fixes (Thanks a lot for that!). Should be good to go! Tested again with unl0kr.
> 
> I added a new commit to properly handle typing with two thumbs by injecting some fake events, with this you can type as fast as you want with your thumbs and not miss an input which is really nice for unl0kr. See the commit description for more details.
> 
> I get that this is a bit of a hack, I'm happy to drop it if it blocks the rest of the series.

The `slot` stuff is clever!

One possible issue I'm seeing is that you're writing `state->slots` from both the polling thread and the LVGL input handler. The locking is avoiding concurrent writes. But depending on how fast events enter and leave the queue, I think it's possible for later multi-touch events to already have overridden the `slots` array filled by earlier ones before you could use it?

I guess generally I'd be nice if LVGL itself supported multi-touch but the event injection strikes me as an acceptable workaround until then (especially given that we already do some event rewriting with `is_relative`).

indev/libinput.c
calebccff: Nice catch! I think you're right.

indev/libinput.c
Johennes: Proposal for fix: https://github.com/calebccff/lv_drivers/pull/4

indev/libinput.c
```diff
@@ -410,50 +550,79 @@ static void reset_scanned_devices(void) {
...
-      pointer_event = libinput_event_get_pointer_event(event);
-      state->most_recent_touch_point.x += libinput_event_pointer_get_dx(pointer_event);
-      state->most_recent_touch_point.y += libinput_event_pointer_get_dy(pointer_event);
-      state->most_recent_touch_point.x = LV_CLAMP(0, state->most_recent_touch_point.x, drv->hor_res - 1);
-      state->most_recent_touch_point.y = LV_CLAMP(0, state->most_recent_touch_point.y, drv->ver_res - 1);
+      evt->point.x += libinput_event_pointer_get_dx(pointer_event);
+      evt->point.y += libinput_event_pointer_get_dy(pointer_event);
+      evt->point.x = LV_CLAMP(0, evt->point.x, drv->hor_res - 1);
+      evt->point.y = LV_CLAMP(0, evt->point.y, drv->ver_res - 1);
+      evt->is_relative = true;
```
Johennes, line 608: One more issue that I just found is that the code currently doesn't support dragging anymore. E.g. if you click and hold but even just slightly move the cursor, that triggers a release because we're not setting `evt->pressed` here.

Unfortunately, `libinput_event_pointer_get_button_state(..)` doesn't seem to work for motion events. So we cannot use it here.

We could maybe define two relativity flags, `is_relative_position` (which would mean add up `x` and `y` to the latest state) and `is_relative_button` (which would mean inherit the `pressed` value from the latest state). `LIBINPUT_EVENT_POINTER_MOTION` events would then require both `is_relative_position` and `is_relative_button` while  `LIBINPUT_EVENT_POINTER_BUTTON` events would only require `is_relative_position`.

Does that make sense?

Johennes:
> One possible issue I'm seeing is that you're writing `state->slots` from both the polling thread and the LVGL input handler. The locking is avoiding concurrent writes. But depending on how fast events enter and leave the queue, I think it's possible for later multi-touch events to already have overridden the `slots` array filled by earlier ones before you could use it?

Actually, I think you can work around that by doing the event injection when reading from libinput as opposed to when consuming the cached events. The `slots` array is guaranteed to only be filled by events preceding the current one at that point.

indev/libinput.c
calebccff: I haven't tested it but I just added a `pointer_button_down` property to `libinput_drv_state_t`, had that attribute be updated on `LIBINPUT_EVENT_POINTER_BUTTON` and set `evt->pressed` based on that. I also made `LIBINPUT_EVENT_POINTER_BUTTON` fallthrough to `LIBINPUT_EVENT_POINTER_MOTION` so that `evt->point` is set.

I think that should work?

Johennes: Nice! Yes, that seems to work. This now actually makes me think we could do the same for the delta movements. Instead of doing the weird `is_relative` thing when consuming events, just store last position we've seen in `state` and add the delta from libinput to it.

Johennes: Have made it so.

Johennes:
> Actually, I think you can work around that by doing the event injection when reading from libinput as opposed to when consuming the cached events. The `slots` array is guaranteed to only be filled by events preceding the current one at that point.

This has been implemented, too, now.

Johennes:
@kisvegabor this is now done from our end and ready for your review. I've tested it correctly in unl0kr with touch, pointer and keyboard devices. Once we make another unl0kr release, I'll start working on porting this into lvgl `master`.

kisvegabor:
Sorry, I've just noticed this PR when the stale bot started to work again after years...

Is it still relevant?

PR #285, opened 2023-08-05
Add USE_WAYLAND to smm.c

cnjhb:
smm.c is not compiled when not using Waylands.
Fix compilation errors,sys/mman not found.


kisvegabor:
Thanks!

PR #286, opened 2023-08-17
Fix crash in sdl.c during startup in certain conditions

mikaeladlers:
I have encountered that the pointer is NULL when I have a lot of screens and objects during startup. If I remove objects I can get it working again which is very strange. So it seams that it is dependent on how many lvObj that has beed created. Adding a null check seams to work.


kisvegabor:
It's really strange. Anyway adding a NULL check is always a valid update . :slightly_smiling_face: 

Thank you!

PR #290, opened 2023-09-01
fix(build): Update xpt2046_read to match signature required by...

### LINKED ISSUE #289: Build of C++ user code fails when attempting to pass xpt2046_read as indev read_cb
LVGL seems to have recently changed the calling convention for an input device driver's read callback. Previously the callback returned `bool` to indicate whether it should be called again; now the callback's return type is `void`, and it sets a flag in place of the old return value.

`xpt2048_read` has not been updated to reflect this change, breaking code that attempts to pass this function as the read callback (warning when compiled as C; error when compiled as C++).

johnauld:
...`lv_indev_drv_t`'s read callback.

The indev read callback previously returned `bool` to indicate whether it should be called again. Per the changelog, this was modified to return this indicator in the `lv_indev_data_t` struct's `continue_reading` member. This change updates the `xpt2046_read` function to be compatible with this new calling convention.

BREAKING CHANGE

Fixes #289.

### DIFF
--- a/indev/XPT2046.c
+++ b/indev/XPT2046.c
@@ -55,9 +55,9 @@ void xpt2046_init(void)
 /**
  * Get the current position and state of the touchpad
  * @param data store the read data here
- * @return false: because no ore data to be read
+ * @return void
  */
-bool xpt2046_read(lv_indev_drv_t * indev_drv, lv_indev_data_t * data)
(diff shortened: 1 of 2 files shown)


kisvegabor:
The drivers in the repo wasn't updated for a while. In LVGL v9 we will add drivers directly into LVGL so they will get more maintenance in the future. 

Anyway, and update here is appreciated, so thank you!

PR #298, opened 2023-10-23
fix(drm): backport drm driver fixes from lvgl 9

niklasf:
Applied fixes/improvements from:

 - https://github.com/lvgl/lvgl/pull/4621
 - https://github.com/lvgl/lvgl/pull/4659
 - https://github.com/lvgl/lvgl/pull/4661

Remains backwards compatible, but replacing `drm_init()` with `drm_disp_drv_init()` is required for optimal performance.

### DIFF
--- a/display/drm.c
+++ b/display/drm.c
@@ -9,19 +9,14 @@
 #include "drm.h"
 #if USE_DRM
 
-#include <unistd.h>
-#include <pthread.h>
-#include <time.h>
-#include <sys/time.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <poll.h>
+#include <stdint.h>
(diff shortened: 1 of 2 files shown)


niklasf:
I just noticed that the `wait_cb` that we can register in LVGL 8 appears to be pretty much exactly what we want to (re)introduce to LVGL 9. So double buffered direct rendering should be possible here, too. I'll make a follow-up pull request for that.

kisvegabor:
> I just noticed that the wait_cb that we can register in LVGL 8 appears to be pretty much exactly what we want to (re)introduce to LVGL 9.

Ahh, good idea! I'm thinking about it.

PR #299, opened 2023-10-24
perf(drm): use wait_cb for direct double-buffered rendering

niklasf:
Follow-up patch mentioned in #298. Slightly improves benchmark performance, as expected.


kisvegabor:
Thanks!

PR #301, opened 2023-11-07
Add pkgconfig file

johanadamnilsson:
Working with lv-drivers headers is a bit problematic. This is because the headers have not the same path when they are build compared to when they are used from another program. This will make it possible to ask `pkg-config --lib` to get required libs and `pkg-config --cflags` to get required cflags.


kisvegabor:
Thank you! As time passes I more and more convinced that having the drivers in the lvgl repo from v9 is a good idea.  It will solve a lot of versioning and synchronization issues.

PR #306, opened 2024-02-12
Windows: create a window which fits selected resolution better

danie1kr:
try to get a window which is closer to the selected resolution. Still 2 pixels off horizontally, but better than what windows creates.

### DIFF
--- a/win32drv/win32drv.c
+++ b/win32drv/win32drv.c
@@ -273,6 +273,9 @@ EXTERN_C HWND lv_win32_create_display_window(
     HICON icon_handle,
     int show_window_mode)
 {
+    RECT rect = { 0, 0, hor_res, ver_res };
+    AdjustWindowRect(&rect, WINDOW_STYLE | WINDOW_EX_STYLE, false);
+    
     HWND display_window_handle = CreateWindowExW(
         WINDOW_EX_STYLE,
         LVGL_SIMULATOR_WINDOW_CLASS,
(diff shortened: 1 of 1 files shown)


kisvegabor:
As it might affect existing projects I wouldn't touch this part.  
However I would be happy to merge https://github.com/lvgl/lvgl/pull/5639 if it's approved by @MouriNaruto.

danie1kr:
Sure, I understand. Actually, I moved from lv_drivers to pure lvgl when I fixed this.

MouriNaruto:
I don't think you have read the implementation carefully. The current 8.x implementation will apply the real size in WM_CREATE message.

https://github.com/lvgl/lv_drivers/blob/451e659cf68d65a7ce13ec7f6a1df16d1c678872/win32drv/win32drv.c#L905-L941

I don't use AdjustWindowRect when creating window because it's not reliable.

image
First, [AdjustWindowRect document](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-adjustwindowrect)  said that to specify an extended window style, use the AdjustWindowRectEx function.

image
Second, [AdjustWindowRectEx document](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-adjustwindowrectex) said that this API is not DPI aware, and should not be used if the calling thread is per-monitor DPI aware. For the DPI-aware version of this API, see AdjustWindowsRectExForDPI.

Also, AdjustWindowsRectExForDPI is not reliable for enabling the DWM composition which is forced beginning with Windows 8. Especially the Windows 10 frameless-style window which is the DWM hacking.

So, this is why I chose only to pass the raw window size before. Because we need to let WM_CREATE message know the raw value for postprocessing. But for 8.x implementation is simplified because 8.x's driver only provides the simulator mode.

Kenji Mouri

PR #307, opened 2024-02-14
fix(wayland): fix compile error wrt LV_WAYLAND_CLIENT_SIDE_DECORATIONS

ThomasDevoogdt:
attach_decoration is called by create_decoration and uses defines that are reserved for decorations

### DIFF
--- a/wayland/wayland.c
+++ b/wayland/wayland.c
@@ -1524,6 +1524,7 @@ static void destroy_graphic_obj(struct graphic_object * obj)
     lv_free(obj);
 }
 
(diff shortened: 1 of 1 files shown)


kisvegabor:
Hi Thomas,

Thank you for the fix! You might have seen that in LVGL v9 we have added the drivers directly to the [LVGL repo](https://github.com/lvgl/lvgl/tree/master/src/drivers). However the Wayland driver is not ported to v9 yet. 

 Maybe do you have time and interest to contribute with that? :slightly_smiling_face:

Issue #287, opened 2023-08-25
Add lvgl v9 support


nicusorcitu:
lv_drivers does not work anylonger with lvgl v9 (due to display rework).
I am trying to build with wayland enabled and got a lot of errors, so I am opening this Issue to track it down.

Is there any roadmap for the fix for lvgl v9 compatibility?

Thanks.
Nicu.

```
In file included from /home/nxf68195/bitbucket/lvgl_mpu/lvgl_imx_demo/lv_drivers/wayland/wayland.h:29,
                 from /home/nxf68195/bitbucket/lvgl_mpu/lvgl_imx_demo/lv_drivers/wayland/wayland.c:9:
/home/nxf68195/bitbucket/lvgl_mpu/lvgl_imx_demo/lvgl/lvgl.h:126:2: warning: #warning "You are using the development version of LVGL which is not stable at this moment. For production use the release/v8.3 branch. To silence this warning add #define LV_USE_DEV_VERSION to lv_conf.h" [-Wcpp]
  126 | #warning "You are using the development version of LVGL which is not stable at this moment. For production use the release/v8.3 branch. To silence this warning add #define LV_USE_DEV_VERSION to lv_conf.h"
      |  ^~~~~~~
In file included from /home/nxf68195/bitbucket/lvgl_mpu/lvgl_imx_demo/lv_drivers/wayland/wayland.c:9:
/home/nxf68195/bitbucket/lvgl_mpu/lvgl_imx_demo/lv_drivers/wayland/wayland.h:33:2: error: #error "Support for user data is required by wayland driver. Set LV_USE_USER_DATA to 1 in lv_conf.h"
   33 | #error "Support for user data is required by wayland driver. Set LV_USE_USER_DATA to 1 in lv_conf.h"
      |  ^~~~~
/home/nxf68195/bitbucket/lvgl_mpu/lvgl_imx_demo/lv_drivers/wayland/wayland.c:196:5: error: unknown type name ‘lv_disp_drv_t’
  196 |     lv_disp_drv_t lv_disp_drv;
      |     ^~~~~~~~~~~~~
/home/nxf68195/bitbucket/lvgl_mpu/lvgl_imx_demo/lv_drivers/wayland/wayland.c:197:5: error: unknown type name ‘lv_disp_draw_buf_t’
  197 |     lv_disp_draw_buf_t lv_disp_draw_buf;
      |     ^~~~~~~~~~~~~~~~~~
[...]

PR #308, opened 2024-03-11
Fix crash when running on Weston 10.0.2 w/pixman rendering backend

etag4048:
Hello,

When running  the wayland client under weston v10.0.2 with the pixman software rendering backend,
The demo application crashes even before the first frame is shown. 

A detailed explanation/investigation is available here https://github.com/etag4048/lv_drivers/blob/weston-pixman-crash-doc/wayland/doc/shared_memory_manager.txt

TL;DR
The shared memory manager introduced in #270 was not designed to work with simple compositors using pixman, it uses a single buffered approach suited for compositors using EGL as a backend, on systems not constrained by memory since it allocates buffers on demand inside the lv_wayland_flush function. 

However, this is not possible on Weston (and probably other more minimalist wayland compositors), because in Weston's case it does not tolerate the resizing of the SHM pool in between two frames.

This PR is only to get the discussion started and is very much a work in progress, as there is still a crash when the window is resized.

A few questions need to be answered.

Is there a need for a double buffered mode for this driver? It seems there are several issues mentioning Weston shipped with Poky from the Yocto project. Or is it intended to be used solely as a LVGL simulator on complex EGL backed compositors ?

While #270  is certainly more flexible, the simple approach taken in the stale pr #253  seems to work best for software rendering, so the solution is to configure shared memory manager to pre-allocated two buffers for the window body. Or should it be ported to the new display API used in LVGL 9 ? as mentioned in issue #287 

Any feedback is greatly appreciated,
Best regards

### DIFF
--- a/wayland/wayland.c
+++ b/wayland/wayland.c
@@ -1158,7 +1158,9 @@ static const struct xdg_toplevel_listener xdg_toplevel_listener = {
     .configure = xdg_toplevel_handle_configure,
     .close = xdg_toplevel_handle_close,
     .configure_bounds = xdg_toplevel_handle_configure_bounds,
-    .wm_capabilities = xdg_toplevel_handle_wm_capabilities
+#ifndef LV_WAYLAND_WESTON_COMPAT
+    .wm_capabilities = xdg_toplevel_handle_wm_capabilities /* This requires newer wayland headers */
+#endif
 };
 
 static void xdg_wm_base_ping(void *data, struct xdg_wm_base *xdg_wm_base, uint32_t serial)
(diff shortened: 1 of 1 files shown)


etag4048:
**This is not a work in progress anymore,** it's a proper fix for the multiple issues that were apparent on Weston with a non-EGL backend when using the shared memory manager introduced in PR#270. This commit changes the way subsurfaces are handled, they are now in 'sync mode' , they are not destroyed on each resize instead another buffer is attached and the position is updated , it also fixes a race condition where XDG configure was called after the the first flush occured. There is also another patch on the way that will make the wayland client compatible with LVGL9

kisvegabor:
It's amazing, thank you very much! It's even better that apparently you have tried it with LVGL v9 too. If so, could you send a  PR for the `lvgl` `master` too?

etag4048:
Thanks, I sure will when it's ready.

Issue #309, opened 2024-03-21
Add SDL high DPI config


slightc:
when I run the simulator on macbook, after reduce the SDL_ZOOM, the screen will become will become extremely blurry, that need add `SDL_WINDOW_ALLOW_HIGHDPI` flag to resolve it when create SDL Window

<img width="480" alt="image" src="https://github.com/lvgl/lv_drivers/assets/17919546/035681e8-1bd4-4db7-8de8-57adc96e825c">

that result of add `SDL_WINDOW_ALLOW_HIGHDPI` flag

<img width="480" alt="image" src="https://github.com/lvgl/lv_drivers/assets/17919546/b1bcd9f4-574e-4029-801b-a2e5c59be96f">

so, I think could add `SDL_HIGHDPI` micro define to config the flag?

Hope for your response, thanks!

PR #310, opened 2024-03-21
feat(sdl): add SDL_HIGHDPI define to allow high DPI render

slightc:
#309


kisvegabor:
What does it change in the practice?

kisvegabor:
The [docs](https://wiki.libsdl.org/SDL2/SDL_WindowFlags) says
> window should be created in high-DPI mode if supported (>= SDL 2.0.1)

So it won't be used if not supported. Therefore I wonder if we can add it unconditionally.

slightc:
I think that should be a config field, it is up to the user to decide on their own

ITotalJustice:
> The [docs](https://wiki.libsdl.org/SDL2/SDL_WindowFlags) says
> 
> > window should be created in high-DPI mode if supported (>= SDL 2.0.1)
> 
> So it won't be used if not supported. Therefore I wonder if we can add it unconditionally.

you mean like?
```c
#if defined(SDL_HIGHDPI) && SDL_HIGHDPI &&  SDL_VERSION_ATLEAST(2, 0, 1)
flag |= SDL_WINDOW_ALLOW_HIGHDPI
#endif
```
to https://github.com/lvgl/lv_drivers/blob/0091dc612facc94dce1061a9b78d641c77f1791a/sdl/sdl.c#L336

slightc:
I think use SDL_HIGHDPI macro definition check, that should be defined by those in need, no need for additional version checking (want to keep simplicity).

slightc:
emm, that like this?
``` c
#ifndef SDL_HIGHDPI
# define SDL_HIGHDPI  SDL_VERSION_ATLEAST(2, 0, 1)
#endif

#if SDL_HIGHDPI
    flag |= SDL_WINDOW_ALLOW_HIGHDPI;
#endif
```

kisvegabor:
Or just:
```c
#if SDL_VERSION_ATLEAST(2, 0, 1)
    flag |= SDL_WINDOW_ALLOW_HIGHDPI;
#endif
```

slightc:
@kisvegabor I'm update it, I think that may best can be force enable/disable by developer, so still use the `SDL_HIGHDPI`, this PR maybe OK?

PR #318, opened 2024-12-07
Filter the sdl driver mouse events to the correct monitor

DDRBoxman:
This change should fix having inputs for multiple displays. This new filtering also allows simulators to add additional non lvgl powered SDL windows that do not pollute the lvgl input data with their mouse presses.

If you set disp on your input driver to NULL it uses the default display (lvgl convention). If you actively set the display field the events will be bound to that SDL window.

In lvgl calling lv_disp_drv_register will automatically create the default display.

Example usage:

```
  lv_disp_t *disp = lv_disp_get_default();

  static lv_indev_drv_t indev_drv_mouse;
  lv_indev_drv_init( &indev_drv_mouse );
  indev_drv_mouse.disp = disp;
  indev_drv_mouse.type = LV_INDEV_TYPE_POINTER;
  indev_drv_mouse.read_cb = sdl_mouse_read;  /*This function will be called periodically (by the library) to get the mouse position and state*/
  lv_indev_drv_register( &indev_drv_mouse );
```

### DIFF
--- a/sdl/sdl.c
+++ b/sdl/sdl.c
@@ -188,6 +188,15 @@ void sdl_display_flush(lv_disp_drv_t * disp_drv, const lv_area_t * area, lv_colo
 
 }
 
+bool sdl_dis_drv_is_monitor_1(lv_disp_drv_t * disp_drv)
+{
+    return disp_drv->flush_cb == sdl_display_flush;
+}
+
+uint32_t monitor_1_win_id(void)
+{
+    return SDL_GetWindowID(monitor.window);
+}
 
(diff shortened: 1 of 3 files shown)


DDRBoxman:
I have the firmware for a TV remote that can be run as a simulator and I wanted to add another SDL window to emulate the hardware keypad from that remote that is not using lvgl code. Unfortunately without this change SDL sends the events for all windows to all the input drivers.

Screenshot 2024-12-06 at 8 09 13 PM

